PR #15, opened 2021-12-29
src: video: prevent size and interval duplicates

### LINKED ISSUE #12: reporting frame size twice
by @williangalvani
image

joaoantoniocardoso:
This commit prevents duplicated sizes and intervals (as reported on #12) from being accepted on video data structures.

Should fix #12.

---

I don't know enough of Rust yet to say if this approach is good, but I can say it works. Anyway, let me know how we can improve it =)

### DIFF
--- a/src/video/video_source_local.rs
+++ b/src/video/video_source_local.rs
@@ -221,6 +221,14 @@ impl VideoSource for VideoSourceLocal {
                     }
                 }
             }
+
+            sizes.sort();
+            sizes.dedup();
+            sizes.iter_mut().for_each(|s| {
+                s.intervals.sort();
+                s.intervals.dedup();
+            });
+
             formats.push(Format {
                 encode: VideoEncodeType::from_str(v4l_format.fourcc.str().unwrap()),
                 sizes,
--- a/src/video/types.rs
+++ b/src/video/types.rs
@@ -11,7 +11,7 @@ pub enum VideoSourceType {
     Local(VideoSourceLocal),
 }
 
-#[derive(Apiv2Schema, Debug, Clone, PartialEq, Deserialize, Serialize)]
+#[derive(Apiv2Schema, Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Deserialize, Serialize)]
 pub enum VideoEncodeType {
     UNKNOWN(String),
     H265,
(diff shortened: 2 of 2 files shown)


src/video/video_source_local.rs
```diff
@@ -221,6 +242,7 @@ impl VideoSource for VideoSourceLocal {
                     }
                 }
             }
+
```
patrickelectric, line 245: Avoid adding style changes on the same commit as function changes. No need to change on this one, but be aware of future commits. `git add -p` or `git checkout -p` is your friend

src/video/video_source_local.rs
```diff
@@ -221,6 +242,7 @@ impl VideoSource for VideoSourceLocal {
                     }
```
patrickelectric, line 242: You could do on the end here a simple:
```
sizes.sort();
sizes.dedup();
sizes.iter_mut().for_each(|size| size.intervals.sort(); size.intervals.dedup());
```

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=4966b30de796a4491abd7b74d61b0a66

src/video/video_source_local.rs
```diff
@@ -207,12 +207,33 @@ impl VideoSource for VideoSourceLocal {
...
-                            })
+                            if sizes
+                                .iter()
+                                .find(|&size| {
+                                    size.width == v4l_size.width && size.height == v4l_size.height
+                                })
+                                .is_none()
+                            {
+                                let mut intervals: Vec<FrameInterval> = vec![];
+                                for v4l_interval in convert_v4l_intervals(enum_frameintervals) {
```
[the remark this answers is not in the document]
joaoantoniocardoso: yes

src/video/video_source_local.rs
joaoantoniocardoso: Nice, my first thought was with `sort().dedup()`, but I wasn't sure if it was OK to change the list order, although the code intent is way clearear. I'll do that!

src/video/video_source_local.rs
joaoantoniocardoso: Sure, I can remove it, I'll be doing a fix commit anyway to change the technique to sort->dedup.

joaoantoniocardoso (commented):
I'll be changing to the sort->dedup technique in the next commits.

patrickelectric (approved):

patrickelectric:
You just need to squash the fixup commit

joaoantoniocardoso:
> You just need to squash the fixup commit

Nice, done!

PR #21, opened 2022-01-06
Use string for usb device description.

### LINKED ISSUE #20: Move UsbBus from struct to string
A struct is problematic since the type of the information can be a tree of usb bus of usb bus.
we are going to change to a string to be more generic.

joaoantoniocardoso:
Because usb descriptions can be way more complex than was expected, we are changing it to be the entire string instead of trying to interpret its parts to populate fields of a struct.

Fix #20.

### DIFF
--- a/src/video/video_source_local.rs
+++ b/src/video/video_source_local.rs
@@ -11,18 +11,11 @@ use v4l::video::Capture;
 
 use log::*;
 
-#[derive(Apiv2Schema, Clone, Debug, PartialEq, Deserialize, Serialize)]
-pub struct UsbBus {
-    pub interface: String,
-    pub usb_hub: u8,
-    pub usb_port: u8,
-}
-
 //TODO: Move to types
 #[derive(Apiv2Schema, Clone, Debug, PartialEq, Deserialize, Serialize)]
 pub enum VideoSourceLocalType {
     Unknown(String),
-    Usb(UsbBus),
+    Usb(String),
     Isp(String),
 }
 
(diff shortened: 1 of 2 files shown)


src/video/video_source_local.rs
```diff
@@ -62,45 +58,31 @@ impl VideoSourceLocalType {
     }
 
     fn usb_from_str(description: &str) -> Option<Self> {
+        let pci_regex_with_hub = Regex::new(
+            r"(?P<interface>[0-9|a-f]+:[0-9|a-f]+:[0-9|a-f]+)\.(?P<usb_hub>\d+-\d)\.(?P<usb_port>\d+)",
+        )
+        .unwrap();
         let pci_regex = Regex::new(
-            r"(?P<interface>[0-9|a-f]+:[0-9|a-f]+:[0-9|a-f]+).(?P<usb_hub>\d+)-(?P<usb_port>\d+)",
+            r"(?P<interface>[0-9|a-f]+:[0-9|a-f]+:[0-9|a-f]+)\.(?P<usb_hub>\d+)-(?P<usb_port>\d+)",
```
patrickelectric, line 66: I would recommend to use the same for pci and pci with hub, but limited until the hub part, for everything else you can do .*

joaoantoniocardoso: @patrickelectric done.

Another option - specially if we were more confident that we are testing all the cases - could be something more descriptive, like:
```Rust
r"^usb-(?P<interface>(([0-9a-fA-F]{2}){1,2}:?){4})\.(usb-)?(?P<usb_hub>(\d[-.]?)+)[-.]+(?P<usb_port>\d[-.]?)+$"
```

patrickelectric:
@joaoantoniocardoso please apply the fixup.

joaoantoniocardoso:
@patrickelectric done.

patrickelectric (approved):

PR #22, opened 2022-01-07
Code cleanup

joaoantoniocardoso:


joaoantoniocardoso:
I can't know for sure, but I'm assuming all the remaining warnings are about dead code that are going to be used in near future.

patrickelectric:
> I can't know for sure, but I'm assuming all the remaining warnings are about dead code that are going to be used in near future.

Since mostly are not used right now, you can add an underscore in the beggning of the name. 
https://doc.rust-lang.org/rustc/lints/listing/warn-by-default.html#explanation-13

rustfmt.toml
```diff
@@ -1 +1 @@
-edition = "2018"
+edition = "2021"
```
patrickelectric, line 1: Can you squash with 99855dc ?

Cargo.lock
```diff
@@ -13,7 +13,7 @@ dependencies = [
  "futures-core",
  "futures-sink",
  "log 0.4.14",
- "pin-project 0.4.28",
+ "pin-project 0.4.29",
```
patrickelectric, line 16: Why this PR is changing the lock file ? This should not happen.

patrickelectric (changes requested):

Cargo.lock
joaoantoniocardoso: oops, I think this had occurred with the `cargo update`.

rustfmt.toml
joaoantoniocardoso: yep. I'll change it.

patrickelectric (approved):

PR #24, opened 2022-01-12
Bring back RTSP

joaoantoniocardoso:


src/stream/stream_backend.rs
```diff
@@ -220,6 +220,75 @@ fn create_udp_stream(
...
 
+fn create_rtsp_stream(
+    video_and_stream_information: &VideoAndStreamInformation,
+) -> Result<StreamType, SimpleError> {
+    let encode = video_and_stream_information
+        .stream_information
+        .configuration
+        .encode
+        .clone();
+    let endpoints = &video_and_stream_information.stream_information.endpoints;
```
patrickelectric, line 231: We should accept a single endpoint for the RTSP server, the reason behind is that RTSP has a single endpoint for factory.
We should also accept only endpoints with the format: `rtsp://localhost/path`, where path is not nested. There is no option to provide RTSP servers outside localhost, what we could accept a port for the RTSP server, but we need to make sure that the same port is used between RTSP streams.

src/stream/stream_backend.rs
```diff
@@ -220,6 +220,75 @@ fn create_udp_stream(
...
+                        "v4l2src device={device}",
+                        " ! video/x-h264,width={width},height={height},framerate={interval_denominator}/{interval_numerator}",
+                    ),
+                    device = &local_device.device_path,
+                    width = configuration.width,
+                    height = configuration.height,
+                    interval_denominator = configuration.frame_interval.denominator,
+                    interval_numerator = configuration.frame_interval.numerator,
+                )
+            } else {
```
patrickelectric, line 251: the if part is already a return, you don't need the else here.

src/stream/stream_backend.rs
```diff
@@ -275,7 +345,43 @@ mod tests {
         assert!(result.is_ok());
         let result = result.unwrap();
 
-        let StreamType::UDP(video_stream_udp) = result;
-        assert_eq!(video_stream_udp.pipeline(), "v4l2src device=/dev/video42 ! video/x-h264,width=1080,height=720,framerate=30/1 ! h264parse ! queue ! rtph264pay config-interval=10 pt=96 ! multiudpsink clients=192.168.0.1:42");
+        let result = match result {
+            StreamType::UDP(video_stream_udp) => video_stream_udp,
+            _any_other_stream_type => panic!("Failed to create UDP stream."),
```
patrickelectric, line 350: `panic!("Wrong stream type for UDP: {:?}", result)`

src/stream/stream_backend.rs
joaoantoniocardoso: Great, thanks for pointing that out!

> We should accept a single endpoint for the RTSP server, the reason behind is that RTSP has a single endpoint for factory.

For now I'll get only the first endpoint from the endpoints vector, could it be done in some better way?

---

> We should also accept only endpoints with the format: `rtsp://localhost/path`, where path is not nested. There is no option to provide RTSP servers outside localhost, what we could accept a port for the RTSP server, but we need to make sure that the same port is used between RTSP streams.

Hmm, to solve that, we could:
- check for rtsp scheme
- check if the domain is localhost
- check if there is no more than 1 path_segment (like segmentA/segmentB/segmengC)

Like so:
```rust
    let endpoint = &video_and_stream_information.stream_information.endpoints[0];
if endpoint.scheme() != "rtsp" {
        return Err(SimpleError::new(format!(
            "The URL's scheme for RTSP endpoints should be \"rtsp\", but was: {:?}",
            endpoint.scheme()
        )));
    } else if endpoint.host_str() != "localhost".into() {
        return Err(SimpleError::new(format!(
            "The URL's host for RTSP endpoints should be \"localhost\", but was: {:?}",
            endpoint.host_str()
        )));
    } else if endpoint.path_segments().iter().count() > 1 {
        return Err(SimpleError::new(format!(
            "The URL's path for RTSP endpoints shouldn't have nested segments (e.g.: \"segmentA/segmentB/segmengC\"), but was: {:?}",
            endpoint.path()
        )));
    }
```

Is there a better topology?

---

Lastly, about the `port`, I'm not sure if we can do it, but maybe we can initialize the RTSP server with the port passed in the first call to create_rtsp_stream, then check if it is the same in the subsequent create_rtsp_stream calls, something like:
```rust
    } else if RTSP_SERVER.is_running() && endpoint.port() != RTSP_SERVER.port() {
        return Err(SimpleError::new(format!(
            "The URL's port for RTSP endpoints should all match to {:?}, but was: {:?}",
            RTSP_SERVER.port(),
            endpoint.port()
        )));
    }
```

src/stream/stream_backend.rs
```diff
@@ -228,11 +228,38 @@ fn create_rtsp_stream(
...
-    let endpoints = &video_and_stream_information.stream_information.endpoints;
+    dbg!(&encode);
+    let endpoint = &video_and_stream_information.stream_information.endpoints[0];
+    dbg!(&endpoint);
+    if endpoint.scheme() != "rtsp" {
+        return Err(SimpleError::new(format!(
+            "The URL's scheme for RTSP endpoints should be \"rtsp\", but was: {:?}",
+            endpoint.scheme()
+        )));
+    } else if endpoint.host_str() != "localhost".into() {
```
patrickelectric, line 239: Else is not necessary after return

patrickelectric (changes requested):

src/stream/stream_backend.rs
joaoantoniocardoso: do you mean like:
```rust
if somethingA { return Err(); }
if somethingB { return Err(); }
if ...
```
instead of:
```rust
if somethingA { return Err(); }
else if somethingB { return Err(); }
else if ...
```
?

Logically the same, indeed cleaner without the elses.

patrickelectric: yes.

src/stream/stream_backend.rs
patrickelectric: To explain my logic here, the `format!` code return an string to the inner scope, but the return is used to the outer scope.
The code is near to the following snippet:
```
fn potato_fun() -> &'static str {
    if 12 * 15 > 150 {
        "Bigger"
    } else {
        "Smaller"
    }
}
// rework
fn potato_fun() -> &'static str {
    if 12 * 15 > 150 {
        "Bigger"
    }

    "Smaller"
}
```

The problem is that the second case is invalid, it appears to be related to the let if pattern: https://doc.rust-lang.org/book/ch03-05-control-flow.html#using-if-in-a-let-statement

src/stream/stream_backend.rs
joaoantoniocardoso: Done (053980b)

joaoantoniocardoso:
@patrickelectric I just rebased and updated the WIP commit.

Current status:
- [x] working with fake streams
- [x] working with camera streams

bugs:
- rtsp server it is only working using port 8554.
- isn't working with multiple streams, ex: when creating two, only the
  first created works.

src/custom/bluerov.rs
```diff
@@ -47,3 +47,49 @@ pub fn udp() -> Vec<VideoAndStreamInformation> {
...
+//                         height: size.height,
+//                         width: size.width,
+//                         frame_interval: size.intervals.first().unwrap().clone(),
+//                     },
+//                 },
+//                 video_source: cam.clone(),
+//             }
+//         })
+//         .collect()
+// }
```
joaoantoniocardoso, line 95: Just keeping it here to ask you if we'll gonna need something like this, 'cause I didn't get the reason why we are creating this default configuration for UDP.

src/stream/_rtsp_server.rs
```diff
@@ -0,0 +1,97 @@
...
+        // the default one.
+        let id = server.attach(None);
+
+        // Start the mainloop. From this point on, the server will start to serve
+        // our quality content to connecting clients.
+        main_loop.run();
+
+        glib::source_remove(id);
+    }
+}
```
joaoantoniocardoso, line 97: Old code, for reference.

src/stream/gst/pipeline_runner.rs
```diff
@@ -8,18 +8,18 @@ pub struct Pipeline {
...
-        }
-    }
-}
+// impl Default for Pipeline {
+//     fn default() -> Self {
+//         Pipeline {
+//             description: //"videotestsrc pattern=ball ! video/x-raw,width=640,height=480 ! videoconvert ! x264enc bitrate=5000 ! video/x-h264, profile=baseline ! rtph264pay ! udpsink host=0.0.0.0 port=5600".into(),
+//         }
+//     }
+// }
```
joaoantoniocardoso, line 22: No generic default string for a pipeline could be used for both UDP and RTSP, I moved it to the UDP/RTSP specific defaults.

src/stream/stream_backend.rs
```diff
@@ -220,6 +220,75 @@ fn create_udp_stream(
     )));
 }
 
+fn create_rtsp_stream(
```
patrickelectric, line 224: these changes are in the stream namespace and not in manager, as the commit message says.

src/stream/rtsp_server.rs
```diff
@@ -0,0 +1,169 @@
...
+            return;
+        }
+
+        loop {
+            std::thread::sleep(std::time::Duration::from_secs(1));
+            if !RTSP_SERVER.as_ref().lock().unwrap().is_running {
+                continue;
+            }
+
+            let mut rtsp_server = RTSP_SERVER.as_ref().lock().unwrap();
```
patrickelectric, line 65: You can do the lock before the if check, using the `rtsp_server` on the if condition.

src/stream/rtsp_server.rs
```diff
@@ -0,0 +1,169 @@
...
+        // provide different streams. Here, we ask our server to give
+        // us a reference to his list of endpoints, so we can add our
+        // test endpoint, providing the pipeline from the cli.
+        let mounts = rtsp_server
+            .server
+            .mount_points()
+            .ok_or("Could not get mount points")
+            .unwrap();
+
+        let factory = &rtsp_server.path_to_factory[path];
```
patrickelectric, line 141: This access does the unwrap for you, we should replace it or we should only run with a previous check to make sure that path exist inside the hashmap.

src/stream/rtsp_server.rs
```diff
@@ -0,0 +1,169 @@
...
+
+        // Now we add a new mount-point and tell the RTSP server to serve the content
+        // provided by the factory we configured above, when a client connects to
+        // this specific path.
+        mounts.add_factory(path, factory);
+
+        if rtsp_server.is_running {
+            return; // main loop thread is already running
```
patrickelectric, line 149: This is unnecessary, we could just set the variable, there is no consequences.

src/stream/rtsp_server.rs
```diff
@@ -0,0 +1,169 @@
...
+        // provide different streams. Here, we ask our server to give
+        // us a reference to his list of endpoints, so we can add our
+        // test endpoint, providing the pipeline from the cli.
+        let mounts = rtsp_server
+            .server
+            .mount_points()
+            .ok_or("Could not get mount points")
+            .unwrap();
+
+        mounts.remove_factory(path);
```
patrickelectric, line 167: You are not removing from the hashmap as well.

src/stream/stream_backend.rs
```diff
@@ -228,49 +229,44 @@ fn create_rtsp_stream(
         .configuration
         .encode
         .clone();
-    dbg!(&encode);
     let endpoint = &video_and_stream_information.stream_information.endpoints[0];
-    dbg!(&endpoint);
```
patrickelectric: This should be squashed with the original commit.

patrickelectric (changes requested):

joaoantoniocardoso (commented):
Thanks for the review, I'll be addressing the code changes by tomorrow.

src/stream/rtsp_server.rs
patrickelectric: I'm not sure if your comment is about this one  or my previous one: https://github.com/bluerobotics/mavlink-camera-manager/pull/24/files#r791807083

src/stream/rtsp_server.rs
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    #[allow(dead_code)]
+    pub fn remove_pipeline(path: &str) {
+        let mut rtsp_server = RTSP_SERVER.as_ref().lock().unwrap();
+        // TODO: handle debug/errors:
+        rtsp_server.path_to_factory.remove(path);
+    }
+
+    pub fn server_start(rtsp_server_host: &str, rtsp_server_port: u16) {
```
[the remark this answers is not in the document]
patrickelectric: Maybe we should change `start` to `start_pipeline`, that would make more change on the RTSPServer perspective.

src/stream/stream_backend.rs
```diff
@@ -220,6 +221,117 @@ fn create_udp_stream(
...
+            "The URL's port for RTSP endpoints should be \"8554\", but was: {:?}",
+            endpoint.port()
+        )));
+    }
+    if endpoint.path_segments().iter().count() > 1 {
+        return Err(SimpleError::new(format!(
+            "The URL's path for RTSP endpoints shouldn't have nested segments (e.g.: \"segmentA/segmentB/segmengC\"), but was: {:?}",
+            endpoint.path()
+        )));
+    }
```
joaoantoniocardoso, line 256: To be fixed: the endpoint mustn't be "" (empty).

src/stream/rtsp_server.rs
```diff
@@ -0,0 +1,169 @@
...
+
+    fn run_main_loop(channel: std::sync::mpsc::Sender<String>) {
+        if let Err(error) = gstreamer::init() {
+            let _ = channel.send(format!("Failed to init GStreamer: {}", error));
+            return;
+        }
+
+        loop {
+            std::thread::sleep(std::time::Duration::from_secs(1));
+            if !RTSP_SERVER.as_ref().lock().unwrap().is_running {
```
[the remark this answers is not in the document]
joaoantoniocardoso: Done, renamed to `run`.

src/stream/rtsp_server.rs
joaoantoniocardoso: Sorry, I was trying to say the following two alternatives:
```rust
        loop {
            std::thread::sleep(std::time::Duration::from_secs(1));

            let mut rtsp_server = RTSP_SERVER.as_ref().lock().unwrap();
            if !rtsp_server.run {
                continue;
            }
```
or
```rust
        loop {
            std::thread::sleep(std::time::Duration::from_secs(1));

            if !RTSPServer::is_running() {
                continue;
            }

            let mut rtsp_server = RTSP_SERVER.as_ref().lock().unwrap();

```

joaoantoniocardoso: The second one is being used in start()

src/stream/video_stream_rtsp.rs
```diff
@@ -0,0 +1,84 @@
+use super::gst::pipeline_runner::Pipeline;
```
patrickelectric, line 1: This commit should not exist, please do the necessary squashing, avoid adding intermediary development commits in master branch.

src/stream/manager.rs
```diff
@@ -36,6 +36,9 @@ pub fn start() {
             StreamType::UDP(stream) => {
                 stream.start();
```
patrickelectric, line 37: I would recommend to change the commit message to only: `src: stream: Add RTSP video stream`
Since you are doing changes in types, manager and backend

src/stream/video_stream_rtsp.rs
```diff
@@ -81,4 +81,4 @@ fn run_video_stream_rtsp(
     channel: std::sync::mpsc::Sender<String>,
 ) {
     unimplemented!();
-}
\ No newline at end of file
+}
```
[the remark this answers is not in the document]
patrickelectric: Ping

src/stream/stream_backend.rs
```diff
@@ -228,7 +228,32 @@ fn create_rtsp_stream(
         .configuration
         .encode
         .clone();
-    let endpoints = &video_and_stream_information.stream_information.endpoints;
+    let endpoint = &video_and_stream_information.stream_information.endpoints[0];
```
patrickelectric, line 232: This should be squashed with the original commit.

src/stream/stream_backend.rs
```diff
@@ -265,7 +266,7 @@ fn create_rtsp_stream(
                 format!(
                     concat!(
                         "v4l2src device={device}",
-                        " ! video/x-h264,width={width},height={height},framerate={interval_denominator}/{interval_numerator}",
+                        " ! video/x-h264,width={width},height={height},framerate={interval_denominator}/{interval_numerator},type=video",
```
patrickelectric, line 269: This should be squashed in the original commit, also this specific commit should be one of the first, since it adds the server source file and the base of the functionality, please correct the commit order.

src/stream/rtsp_server.rs
```diff
@@ -156,5 +156,9 @@ impl RTSPServer {
             .unwrap();
 
         mounts.remove_factory(path);
+
+        rtsp_server.path_to_factory.remove(path);
```
patrickelectric, line 160: again, please squash.

patrickelectric (changes requested):
The code looks correct, but the commits needs to be organized and squashed correctly, avoid adding unnecessary changes in code that is not merged on master, avoid also creating unnecessary intermediary commits for code that was not merged, start commiting the "base" and "heavy" part of the code and after that commits that add such functionality on the existent codebase.

patrickelectric (approved):
Perfect, thanks !

Issue #38, opened 2022-01-27
MAVlink information needs to be updated to support both RTSP and RTPUDP depending of the video stream 


patrickelectric:
https://github.com/patrickelectric/mavlink-camera-manager/blob/ed527463feef50617012ec88106a04035542be86/src/mavlink/mavlink_camera.rs#L537

joaoantoniocardoso:
https://mavlink.io/en/messages/common.html#VIDEO_STREAM_TYPE

PR #39, opened 2022-01-27
src: video,stream: Set mavtype according to the video stream type

joaoantoniocardoso:
Fixes https://github.com/bluerobotics/mavlink-camera-manager/issues/38.

### DIFF
--- a/src/mavlink/mavlink_camera.rs
+++ b/src/mavlink/mavlink_camera.rs
@@ -32,6 +32,7 @@ pub struct MavlinkCameraComponent {
 pub struct MavlinkCameraInformation {
     component: MavlinkCameraComponent,
     mavlink_connection_string: String,
+    mavlink_stream_type: mavlink::common::VideoStreamType,
     video_stream_uri: Url,
     video_source_type: VideoSourceType,
     vehicle: Arc<Box<dyn mavlink::MavConnection<mavlink::common::MavMessage> + Sync + Send>>,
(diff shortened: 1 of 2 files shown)


src/mavlink/mavlink_camera.rs
```diff
@@ -534,7 +548,7 @@ fn video_stream_information(video_name: &str, video_uri: &str) -> mavlink::commo
             hfov: 0,
             stream_id: 1, // Starts at 1, 0 is for broadcast
             count: 0,
-            mavtype: mavlink::common::VideoStreamType::VIDEO_STREAM_TYPE_RTPUDP,
```
patrickelectric, line 537: We can extract the video stream type based on the video url scheme, that'll make the patch much simpler. 
The schemes that the url can assume are:
```
"tsusb://" -> udp h264
"udp://" -> h264
"rtsp://"
"udp265://" -> udp h265
"tcp://" -> MPEG
"mpegts://"
```
But the protocol only supports RTSP and UDP for now.

joaoantoniocardoso: I see, but wouldn't it be harder to maintain if we keep several parsers in different parts of the code?

patrickelectric: They are mavlink/vehicle specific strings, but I understand your point.

patrickelectric (approved):

PR #41, opened 2022-01-27
Add thermal camera flag

### LINKED ISSUE #27: Add option to set camera as thermal

joaoantoniocardoso:
Fixes #27.

### DIFF
--- a/src/mavlink/mavlink_camera.rs
+++ b/src/mavlink/mavlink_camera.rs
@@ -35,6 +35,7 @@ pub struct MavlinkCameraInformation {
     mavlink_stream_type: mavlink::common::VideoStreamType,
     video_stream_uri: Url,
     video_source_type: VideoSourceType,
+    thermal: bool,
     vehicle: Arc<Box<dyn mavlink::MavConnection<mavlink::common::MavMessage> + Sync + Send>>,
 }
 
@@ -64,6 +65,7 @@ impl std::fmt::Debug for MavlinkCameraInformation {
             .field("mavlink_stream_type", &self.mavlink_stream_type)
             .field("video_stream_uri", &self.video_stream_uri)
             .field("video_source_type", &self.video_source_type)
+            .field("thermal", &self.thermal)
             .finish()
     }
 }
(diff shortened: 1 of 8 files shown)


src/mavlink/mavlink_camera.rs
```diff
@@ -35,6 +35,7 @@ pub struct MavlinkCameraInformation {
     mavlink_stream_type: mavlink::common::VideoStreamType,
     video_stream_uri: Url,
     video_source_type: VideoSourceType,
+    thermal: bool,
```
patrickelectric, line 38: This should be an optional.

src/video/types.rs
```diff
@@ -26,6 +26,7 @@ pub struct CaptureConfiguration {
     pub height: u32,
     pub width: u32,
     pub frame_interval: FrameInterval,
+    pub thermal: bool,
```
patrickelectric, line 29: I think that we should expose such variable in a different way, maybe via a vector of enum of capabilities, or a struct.
The reason behind that is that we could add more fields, making the structure a bit confuse, also this is not a capture configuration option (as in the gstreamer pipeline), but as a camera capability thing. 
Maybe storing this in a total different structure may makes more sense, or a structure that encapsulates both.

src/stream/types.rs
```diff
@@ -34,10 +36,30 @@ impl StreamType {
 impl Apiv2Schema for Url {
 }*/
 
+#[derive(Apiv2Schema, Clone, Debug, PartialEq, Deserialize, Serialize)]
+pub struct CaptureConfiguration {
```
patrickelectric, line 40: Why is this in stream and not in video ?

joaoantoniocardoso: Hmm, may you tell me, there was a "TODO" comment in the video saying to move it to the stream;

patrickelectric: Ook so :laughing:

patrickelectric (approved):

src/mavlink/mavlink_camera.rs
joaoantoniocardoso: Just to document, it could, but because the ExtendedConfiguration struct implements the Default trait, we'll always have a value available for it.

Issue #50, opened 2022-02-10
Move supported schemes to ENUM


patrickelectric:
The schemes that we are using, such as: "udp" | "udp265"| "rtsp" | "mpegts" | "tcp", should be enums to deal with it in a safer way

PR #60, opened 2022-03-11
Add webrtc

### LINKED ISSUE #53: Add webrtc server
https://rustrepo.com/repo/centricular-webrtcsink-rust-network-programming

joaoantoniocardoso:
Fixes #53.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,7 +37,7 @@ sys-info = "0.9.1"
 chrono = "0.4.19"
 env_logger = "0.9.0"
 lazy_static = "1.4.0"
-log = "0.4.14"
+log = "0.4.16"
 paperclip = { version = "0.6.1", features = ["actix", "swagger-ui", "url"] }
 #TODO: Replace it with yaserde
 quick-xml = { version = "0.22.0", features = ["serialize"] }
@@ -55,10 +55,14 @@ glib = { version = "0.15.2", optional = true }
 gstreamer = { version = "0.18.1", optional = true }
 gstreamer-rtsp-server = { version = "0.18.0", optional = true }
 
-## WebRTC Turn server
-tokio = { version = "1.15.0", features = ["full"] }
-turn = "0.5.4"
-util = { package = "webrtc-util", version = "0.5.3", default-features = false, features = ["conn", "vnet"] }
+## WebRTC Turn/Stun and Signalling server
+anyhow = "1"
+async-std = "1"
+async-native-tls = "0.4"
+tokio = "1.17"
+turn = "0.5"
+util = { package = "webrtc-util", version = "0.5", default-features = false, features = ["vnet"] }
+webrtcsink-signalling = { version = "0.1.0", git = "https://github.com/centricular/webrtcsink", rev = "55d30db53bb3931f6477b6c1bad4de2a5ec5f7e4" }
 
 [dev-dependencies]
 rand = "0.8.4"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,10 +50,15 @@ pnet = "0.29.0"
 ## Mavlink
 mavlink = { version = "0.10.0", features = ["default", "emit-extensions"] }
 
-## RTSP server
+## GSTREAMER
 glib = { version = "0.15.2", optional = true }
 gstreamer = { version = "0.18.1", optional = true }
 gstreamer-rtsp-server = { version = "0.18.0", optional = true }
+
+## WebRTC Turn server
+tokio = { version = "1.15.0", features = ["full"] }
+turn = "0.5.4"
+util = { package = "webrtc-util", version = "0.5.3", default-features = false, features = ["conn", "vnet"] }
 
 [dev-dependencies]
 rand = "0.8.4"
--- a/src/server/manager.rs
+++ b/src/server/manager.rs
@@ -50,7 +50,10 @@ pub fn run(server_address: &str) {
             // Record services and routes for paperclip OpenAPI plugin for Actix.
             .data(web::JsonConfig::default().error_handler(json_error_handler))
             .route("/", web::get().to(pages::root))
-            .route(r"/{filename:.*(\.html|\.js)}", web::get().to(pages::root))
+            .route(
+                r"/{filename:.*(\.html|\.js|\.css)}",
+                web::get().to(pages::root),
+            )
             .route("/delete_stream", web::delete().to(pages::remove_stream))
             .route("/streams", web::get().to(pages::streams))
             .route("/streams", web::post().to(pages::streams_post))
--- a/src/server/manager.rs
+++ b/src/server/manager.rs
@@ -1,4 +1,5 @@
 use super::pages;
+use crate::stream::signalling_server::SignallingServer;
 
 use actix_web::{
     error::{ErrorBadRequest, JsonPayloadError},
@@ -22,6 +23,9 @@ fn json_error_handler(error: JsonPayloadError, _: &HttpRequest) -> actix_web::Er
 
 // Start REST API server with the desired address
 pub fn run(server_address: &str) {
+    // Start WebRTC signalling server before the HTTP so it can answer any request comming from the http front-end.
+    SignallingServer::start();
+
     let server_address = server_address.to_string();
 
     // Start HTTP server thread
--- a/src/server/pages.rs
+++ b/src/server/pages.rs
@@ -64,6 +64,8 @@ pub fn load_file(file_name: &str) -> String {
     match file_name {
         "" | "index.html" => std::include_str!("../html/index.html").into(),
         "vue.js" => std::include_str!("../html/vue.js").into(),
+        "webrtc/index.html" => std::include_str!("../html/webrtc/index.html").into(),
+        "webrtc/webrtc.js" => std::include_str!("../html/webrtc/webrtc.js").into(),
         _ => format!("File not found: {}", file_name),
     }
 }
@@ -73,6 +75,8 @@ pub fn root(req: HttpRequest) -> HttpResponse {
     let path = match filename {
         "" | "index.html" => load_file("index.html"),
         "vue.js" => load_file("vue.js"),
+        "webrtc/" | "webrtc/index.html" => load_file("webrtc/index.html"),
+        "webrtc/webrtc.js" => load_file("webrtc/webrtc.js"),
         something => {
             //TODO: do that in load_file
             return HttpResponse::NotFound()
--- a/src/server/pages.rs
+++ b/src/server/pages.rs
@@ -69,7 +69,8 @@ pub fn load_file(file_name: &str) -> String {
 }
 
 pub fn root(req: HttpRequest) -> HttpResponse {
-    let path = match req.match_info().query("filename") {
+    let filename = req.match_info().query("filename");
+    let path = match filename {
         "" | "index.html" => load_file("index.html"),
         "vue.js" => load_file("vue.js"),
         something => {
@@ -79,7 +80,15 @@ pub fn root(req: HttpRequest) -> HttpResponse {
                 .body(format!("Page does not exist: {}", something));
         }
     };
-    HttpResponse::Ok().content_type("text/html").body(path)
+    if filename.ends_with(".js") {
+        return HttpResponse::Ok()
+            .content_type("text/javascript")
+            .body(path);
+    }
+    if filename.ends_with(".css") {
+        return HttpResponse::Ok().content_type("text/css").body(path);
+    }
+    return HttpResponse::Ok().content_type("text/html").body(path);
 }
 
 //TODO: change endpoint name to sources
--- a/src/stream/manager.rs
+++ b/src/stream/manager.rs
@@ -39,6 +39,9 @@ pub fn start() {
             StreamType::RTSP(stream) => {
                 stream.start();
             }
+            StreamType::WEBRTC(stream) => {
+                stream.start();
+            }
             StreamType::REDIRECT(_) => (),
         }
     }
@@ -87,6 +90,10 @@ pub fn add_stream_and_start(
             "tcp" => mavlink::common::VideoStreamType::VIDEO_STREAM_TYPE_TCP_MPEG,
             "udp" | _ => mavlink::common::VideoStreamType::VIDEO_STREAM_TYPE_RTPUDP,
         },
+        // TODO: update WEBRTC arm with the correct type once mavlink starts to support it.
+        // Note: For now this is fine because most of the clients doesn't seems to be using mavtype to determine the stream type,
+        // instead, they're parsing the URI's scheme itself, so as long as we pass a known scheme, it should be enough.
+        StreamType::WEBRTC(_) => mavlink::common::VideoStreamType::VIDEO_STREAM_TYPE_RTSP,
     };
 
     stream.mut_inner().start();
--- a/src/stream/manager.rs
+++ b/src/stream/manager.rs
@@ -68,9 +68,11 @@ pub fn add_stream_and_start(
     let mut manager = MANAGER.as_ref().lock().unwrap();
 
     for stream in manager.streams.iter() {
-        stream
-            .video_and_stream_information
-            .conflicts_with(&video_and_stream_information)?
+        if !stream.stream_type.inner().allow_same_endpoints() {
+            stream
+                .video_and_stream_information
+                .conflicts_with(&video_and_stream_information)?
+        }
     }
 
     let endpoint = video_and_stream_information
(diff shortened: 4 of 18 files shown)


src/video_stream/types.rs
```diff
@@ -38,7 +38,7 @@ impl VideoAndStreamInformation {
         let other_endpoints: HashSet<_> = other.stream_information.endpoints.iter().collect();
         let common_endpoints: HashSet<_> = our_endpoints.intersection(&other_endpoints).collect();
 
-        if !common_endpoints.is_empty() {
+        if !common_endpoints.is_empty() && self.stream_information.endpoints[0].scheme() != "turn" {
```
patrickelectric, line 41: We should add a `allow_same_endpoint` bool for webrtc stream type. and avoid this check here.

patrickelectric (changes requested):

src/stream/turn_server.rs
```diff
@@ -0,0 +1,216 @@
...
+        let mut turn_server = TURN_SERVER.as_ref().lock().unwrap();
+        turn_server.run = true;
+        turn_server.running_peers += 1;
+    }
+
+    pub fn notify_stop() {
+        debug!("TURN server was notified with a Stop!");
+        let mut turn_server = TURN_SERVER.as_ref().lock().unwrap();
+
+        if turn_server.running_peers != 0 {
```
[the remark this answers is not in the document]
patrickelectric: it's unsigned.

src/stream/turn_server.rs
```diff
@@ -0,0 +1,216 @@
...
+            turn_server.running_peers -= 1;
+        }
+
+        if turn_server.running_peers == 0 {
```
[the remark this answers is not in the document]
patrickelectric: it's unsigned.

src/stream/manager.rs
```diff
@@ -78,31 +81,41 @@ pub fn add_stream_and_start(
...
-        },
-    };
 
-    stream.mut_inner().start();
-    manager.streams.push(Stream {
-        stream_type: stream,
-        video_and_stream_information: video_and_stream_information.clone(),
-        mavlink_camera: MavlinkCameraHandle::new(
+    let mavlink_camera_handle = if let StreamType::WEBRTC(_) = &stream {
+        None
```
patrickelectric, line 88: You can use `webrtc://<ice-user>:<ice-password>@<Address/Port>?<parameters>` for the video uri in mavlink manager.

src/stream/stream_backend.rs
```diff
@@ -136,6 +137,26 @@ fn check_scheme(
                     return Err(SimpleError::new(format!("Endpoint with udp265 scheme only supports H265 encode. Encode: {:?}, Endpoints: {:#?}", encode, endpoints)));
                 }
             }
+            "turn" => {
```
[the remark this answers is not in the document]
patrickelectric: @joaoantoniocardoso what was the conclusion of that ?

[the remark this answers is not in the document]
patrickelectric: Oh, that the user could use turn or stun, right ?

src/stream/stream_backend.rs
```diff
@@ -370,6 +387,109 @@ fn create_redirect_stream(
...
+    }
+
+    let configuration = match &video_and_stream_information
+        .stream_information
+        .configuration
+    {
+        CaptureConfiguration::VIDEO(configuration) => configuration,
+        _ => return Err(SimpleError::new(format!(
+            "The only accepted configuration type is the VideoCaptureConfiguration, which is like {supported_configuration} ",
+            supported_configuration="VideoCaptureConfiguration { \"encode\": <string>, \"height\": <integer>, \"width\": <integer>, \"frame_interval\": <integer/integer> }",
```
patrickelectric, line 413: Isn't a better way of doing it ?

src/stream/stream_backend.rs
```diff
@@ -370,6 +387,109 @@ fn create_redirect_stream(
...
+            "The only accepted endpoint is: {expected_endpoint}, but {endpoint} was given."
+        )));
+    }
+
+    let configuration = match &video_and_stream_information
+        .stream_information
+        .configuration
+    {
+        CaptureConfiguration::VIDEO(configuration) => configuration,
+        _ => return Err(SimpleError::new(format!(
```
patrickelectric, line 411: Can you print what was received ?

src/stream/stream_backend.rs
```diff
@@ -370,6 +387,109 @@ fn create_redirect_stream(
...
+
+    let encode = configuration.encode.clone();
+    let video_source = &video_and_stream_information.video_source;
+    let pipeline = match video_source {
+        VideoSourceType::Local(local_device) => {
+            if VideoEncodeType::H264 == encode {
+                format!(
+                    concat!(
+                        "v4l2src device={device}",
+                        " ! video/x-h264,width={width},height={height},framerate={interval_denominator}/{interval_numerator}",
```
patrickelectric, line 425: Create the caps in a previous line and use it here to avoid duplicated lines.

src/stream/stream_backend.rs
```diff
@@ -386,6 +506,7 @@ fn create_stream(
         match endpoint.scheme() {
             "udp" => create_udp_stream(video_and_stream_information),
             "rtsp" => create_rtsp_stream(video_and_stream_information),
+            "turn" => create_webrtc_stream(video_and_stream_information),
```
patrickelectric, line 510: `create_webrtc_turn_stream`

src/stream/turn_server.rs
```diff
@@ -0,0 +1,216 @@
...
+            Err(Error::ErrFakeErr)
+        }
+    }
+}
+
+lazy_static! {
+    pub static ref TURN_SERVER: Arc<Mutex<TurnServer>> =
+        Arc::new(Mutex::new(TurnServer::default()));
+}
+
```
patrickelectric, line 62: We should have a public function to access the url of the turn server, with that we can avoid the hardcoded url around the source code for it.

src/stream/manager.rs
```diff
@@ -136,6 +136,7 @@ pub fn remove_stream(stream_name: &str) -> Result<(), SimpleError> {
     let mut manager = MANAGER.as_ref().lock().unwrap();
     match manager.streams.iter().position(find_stream) {
         Some(index) => {
+            manager.streams[index].stream_type.mut_inner().stop();
```
patrickelectric, line 139: That's a bit weird, why the stream can't stop itself on the desconstructor ?

src/server/manager.rs
```diff
@@ -22,6 +23,8 @@ fn json_error_handler(error: JsonPayloadError, _: &HttpRequest) -> actix_web::Er
 
 // Start REST API server with the desired address
 pub fn run(server_address: &str) {
+    SignallingServer::start();
```
patrickelectric, line 26: Can you explain a bit better why it should start before the http server ?

patrickelectric (changes requested):
The CI is also failing

src/stream/stream_backend.rs
joaoantoniocardoso: Probably... but not for cheap. I'd asked about it in the BR rust community and it seems that the only solution to make this automatic would be to create a specific macro for that. I'd tried and it feels just too much effort for now... My approach was basically to create a macro that adds a custom implementation of a function that prints <name>: <type> of each field, but to use it, we would need to wrap all structs with that macro - that's ugly.

src/stream/manager.rs
joaoantoniocardoso: How would the stream know that it needs to be deconstructed once the client has clicked to 'remove' the stream? That's why I needed to add this here, and to me, it seems coherent with the `start()` called in the `add_stream_and_start()`.

src/server/manager.rs
joaoantoniocardoso: I just thought that if the signaling server started _after_ the http, the frontend could request the signaling server before it's ready - which it's not a big deal, but can be avoided.

src/stream/stream_backend.rs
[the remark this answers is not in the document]
joaoantoniocardoso: Supposedly yes, but I hadn't implemented that yet. The idea is that the user could pass an arbitrary stun and/or turn servers to be used instead of our local one. Now we have the signalling server too, which, in theory, could also be created by the user if it uses the [`protocol` specified here](https://github.com/centricular/webrtcsink/tree/main/protocol).

To implement this behavior we would need to use our local default unless it's specified by the user, and also identify as a webrtc stream for if a `stun://`, `turn://` or `ws://` is present in the list of endpoints. What do you think?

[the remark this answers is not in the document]
joaoantoniocardoso: Okay, I did the changes, take a look now - the client can just use `webrtc://` (without host/ip) to use our local servers, or he can specify arbitrary turn, stun, and signalling servers to be used instead.

src/stream/manager.rs
joaoantoniocardoso: I didn't get the idea, I'd looked for any webrtc thing in mavlink and couldn't find anything, can you provide a reference of that?

joaoantoniocardoso:
> The CI is also failing

yes, I'm not sure if we can publish our project using a dependency that is not published yet (webrtcsink). Perhaps we'd ask them about release plans..?

---

What was tested:
- WebRTC on Edge (Linux), Chrome (Linux, Android), and Firefox (Linux, Android).
- our default TURN/STUN/SIGNALLING servers.
- custom remote TURN/STUN servers.

What does not work:
- WebRTC without an interface - this is a browser limitation. I haven't tested it with a local network (between desktop and raspberry), but it should work.

What needs to be done:
- the client html/js. (I'd need help with the routing stuff in server/manager.rs)
- test with raspberry.

src/stream/manager.rs
patrickelectric: When you remove something from the list, the item will not have a owner and will call the constructor, there is no need to call stop manually, it should be done in the desconstructor.

src/server/manager.rs
patrickelectric: So, can you add a comment about it ?

Cargo.toml
```diff
@@ -50,11 +50,20 @@ pnet = "0.29.0"
...
+## GSTREAMER
 glib = { version = "0.15.2", optional = true }
 gstreamer = { version = "0.18.1", optional = true }
 gstreamer-rtsp-server = { version = "0.18.0", optional = true }
 
+## WebRTC Turn/Stun and Signalling server
+anyhow = "1"
+async-std = { version = "1", features = ["unstable", "attributes"] }
+async-native-tls = "0.3"
+tokio = { version = "1.17.0", features = ["full"] }
```
patrickelectric, line 62: just to be sure, do we need to have full features ?

src/stream/manager.rs
joaoantoniocardoso: Hmm, interesting design, I don't know why, but it is not how it is working. I'll try to debug it.

src/stream/stream_backend.rs
```diff
@@ -136,6 +137,26 @@ fn check_scheme(
                     return Err(SimpleError::new(format!("Endpoint with udp265 scheme only supports H265 encode. Encode: {:?}, Endpoints: {:#?}", encode, endpoints)));
                 }
             }
+            "turn" => {
+                if VideoEncodeType::H264 != encode {
+                    return Err(SimpleError::new(format!("Endpoint with WebRTC scheme only supports H264 encode. Encode: {:?}, Endpoints: {:#?}", encode, endpoints)));
+                }
+
+                //WebRTC endpoints should contain user, credential, host and port
```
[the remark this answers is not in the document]
joaoantoniocardoso: Now the front-end can simply pass `webrtc://` as the endpoint.

src/stream/stream_backend.rs
```diff
@@ -370,6 +391,117 @@ fn create_redirect_stream(
...
+    video_and_stream_information: &VideoAndStreamInformation,
+) -> Result<StreamType, SimpleError> {
+    let endpoint = &video_and_stream_information.stream_information.endpoints[0];
+    if endpoint.scheme() != "turn" {
+        return Err(SimpleError::new(format!(
+            "The URL's scheme for WebRTC endpoints should be \"turn\", but was: {:?}",
+            endpoint.scheme()
+        )));
+    }
+    if endpoint.host_str() != "0.0.0.0".into() {
```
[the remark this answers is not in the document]
joaoantoniocardoso: Thanks, the logic has changed entirely, so this is not relevant anymore.

src/stream/manager.rs
patrickelectric: You can add an url for the mavlink camera manager and not use none, having a mavlink camera handle for the webrtc stream.

src/stream/stream_backend.rs
```diff
@@ -37,10 +41,14 @@ fn check_endpoints(
         return Err(SimpleError::new("Endpoints are empty".to_string()));
     }
 
+    let is_custom_webrtc = (endpoints[0].scheme() == "stun")
+        || (endpoints[0].scheme() == "turn")
+        || (endpoints[0].scheme() == "ws");
```
patrickelectric, line 46: ws stands for websocket and for not webrtc, is this correct ? should we use "webrtc" ?

src/stream/stream_backend.rs
```diff
@@ -136,6 +144,23 @@ fn check_scheme(
                     return Err(SimpleError::new(format!("Endpoint with udp265 scheme only supports H265 encode. Encode: {:?}, Endpoints: {:#?}", encode, endpoints)));
                 }
             }
+            "webrtc" | "stun" | "turn" | "ws" => {
```
patrickelectric, line 147: Check my previous comment https://github.com/bluerobotics/mavlink-camera-manager/pull/60/files#r836481813

patrickelectric, line 147: We seriously need to handle that without the need of strings. We can do that in another PR.

src/stream/stream_backend.rs
patrickelectric: It would be helpful that serde could provide that for us, but we can take a better look on the future.

patrickelectric:
> Perhaps we should ask them about release plans..?

We should at least disable the deployment check step for now.

src/stream/stream_backend.rs
patrickelectric: You can also do != webrtc

src/stream/manager.rs
joaoantoniocardoso: Perfect, I'd forgotten about implementing Drop() to call stop().

Cargo.toml
joaoantoniocardoso: we needn't

Cargo.toml
```diff
@@ -50,11 +50,20 @@ pnet = "0.29.0"
...
 
-## RTSP server
+## GSTREAMER
 glib = { version = "0.15.2", optional = true }
 gstreamer = { version = "0.18.1", optional = true }
 gstreamer-rtsp-server = { version = "0.18.0", optional = true }
 
+## WebRTC Turn/Stun and Signalling server
+anyhow = "1"
+async-std = { version = "1", features = ["unstable", "attributes"] }
```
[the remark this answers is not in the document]
joaoantoniocardoso: you're right, thanks

src/stream/stream_backend.rs
joaoantoniocardoso: indeed, and we already have plans to do that: https://github.com/bluerobotics/mavlink-camera-manager/issues/50

src/server/pages.rs
```diff
@@ -64,22 +64,43 @@ pub fn load_file(file_name: &str) -> String {
     match file_name {
         "" | "index.html" => std::include_str!("../html/index.html").into(),
         "vue.js" => std::include_str!("../html/vue.js").into(),
+        "webrtc/index.html" => std::include_str!("../html/webrtc/index.html").into(),
```
patrickelectric, line 67: Be sure to remove all files that are incompatible with the MIT license, we can't include it at compile time.

patrickelectric, line 67: This function also needs to be updated to load files at runtime that is not the cargo folder, since we are going to download the webrtc files that are incompatible with our license.

src/stream/manager.rs
joaoantoniocardoso: Oh, I still need to do this part. Because webrtc definition is a combination of stun, turn, and signalling servers, how are we going to send it? I haven't checked yet, but we might only need the signalling server to be able to receive the stream somewhere else.

patrickelectric: We could send only one of them, the first one provided.

patrickelectric (approved):

PR #64, opened 2022-04-11
Load webrtc resources in runtime

joaoantoniocardoso:

PR #65, opened 2022-04-18
Webrtc check if available

### LINKED ISSUE #61: Add check and warning message for missing webrtcsink
Also disable webrtcsink pipeline creation if the element is not avaiable

joaoantoniocardoso:
This PR should close #61.


patrickelectric:
> This PR should close #61.
> 
> DRAFT because it should be rebased after #64. To review it, only the two last commits matter:
> 
> * _src: stream: Check if webrtcsink is available, return error if not._
> * _src: server: Check if webrtcsink is available, show a warning if not._

Is there a reason for this to be based over #64 and not over master ?

joaoantoniocardoso:
> > This PR should close #61.
> > DRAFT because it should be rebased after #64. To review it, only the two last commits matter:
> > 
> > * _src: stream: Check if webrtcsink is available, return error if not._
> > * _src: server: Check if webrtcsink is available, show a warning if not._
> 
> Is there a reason for this to be based over #64 and not over master ?

yes, it was made on top of [this](https://github.com/bluerobotics/mavlink-camera-manager/pull/65/files#diff-700b33aeb6bb8e8bc5faddcaaea53ac801451ff4a90eb7ca659c1e8415e30e15R43-R63) block introduced by #64

patrickelectric:
@joaoantoniocardoso please rebase over master

src/stream/turn_server.rs
```diff
@@ -15,6 +15,9 @@ use turn::Error;
 
 use log::*;
 
+pub const WEBRTCSINK_PLUGIN_NAME: &str = "webrtcsink";
+pub const WEBRTCSINK_VERSION: &str = "0.1.0";
```
patrickelectric, line 19: this webrtc specific stuff should be in a webrtc file, if turn server is webrtc only, for organization, we should move the necessary files in a stream/webrtc folder.

src/stream/stream_backend.rs
```diff
@@ -398,6 +420,18 @@ fn create_redirect_stream(
 fn create_webrtc_turn_stream(
     video_and_stream_information: &VideoAndStreamInformation,
 ) -> Result<StreamType, SimpleError> {
+    if !is_gstreamer_plugin_available(WEBRTCSINK_PLUGIN_NAME, WEBRTCSINK_VERSION) {
```
patrickelectric, line 423: we should add on the same place as the webrtc file a test like:
```
`is_webrtcsink_available() -> bool`
```
Where it'll print the error message.

src/stream/stream_backend.rs
```diff
@@ -32,6 +34,26 @@ pub fn new(
     return create_stream(video_and_stream_information);
 }
 
+pub fn is_gstreamer_plugin_available(plugin_name: &str, min_version: &str) -> bool {
```
patrickelectric, line 37: Why is this here and not in the gst folder ?

patrickelectric (changes requested):

patrickelectric (commented):
It looks good, I'll test it soon.

src/stream/gst/utils.rs
```diff
@@ -0,0 +1,19 @@
...
+    let mut version = [0, 0, 0];
+    version
+        .iter_mut()
+        .zip(min_version.split('.').map(|s| s.parse().unwrap_or(0u32)))
+        .map(|(a, b)| *a = b)
+        .for_each(drop);
+
+    let registry = gstreamer::Registry::get();
+    registry.lookup_feature(plugin_name).is_some()
+        && registry.check_feature_version(plugin_name, version[0], version[1], version[2])
```
patrickelectric, line 18:
```suggestion
    use semver::Version;
    
    let version = Version::from_str(min_version).unwrap();
    let registry = gstreamer::Registry::get();
    registry.lookup_feature(plugin_name).is_some()
        && registry.check_feature_version(
            plugin_name,
            version.major, version.minor, version.patch)
```

Also, a question, what happens if you use check_feature_version directly without lookup feature ? 
It would be helpful to check for feature first and after that check for version and have two different messages to inform the user about the problem. 
With the current code, I don't see the first check necessary.

patrickelectric (changes requested):

src/stream/gst/utils.rs
joaoantoniocardoso: Thanks for the semver suggestion, much cleaner that way.

About the lookup_feature, I'd asked myself the same question, and I have tested that hypothesis and it works, but because the source code for gst-inspect was doing exactly that and I don't know if there are edge cases that have motivated them to double-check that, I've decided to just repeat them instead of investigating it further. Should we really care about it?

joaoantoniocardoso: Okay, now I double-checked and it seems to be used because the APIs are, in fact, different, and in C, they are getting a GstPluginFeature to use in gst_plugin_feature_check_version, which is not needed in the rust API.

patrickelectric:
@joaoantoniocardoso appears to have a problem with the style :laughing:

patrickelectric (approved):
LGTM
image

PR #67, opened 2022-04-21
Fix PARAM_EXT_SET

joaoantoniocardoso:
This adds the basic ability to have camera controls set by QGC.

### DIFF
--- a/src/mavlink/mavlink_camera.rs
+++ b/src/mavlink/mavlink_camera.rs
@@ -373,15 +373,42 @@ fn receive_message_loop(
                             error!("Failed to parse control id: {:#?}", error);
                             continue;
                         }
-                        let _control_id = control_id.unwrap();
+                        let control_id = control_id.unwrap();
 
                         if let Err(error) = control_value {
                             error!("Failed to parse parameter value: {:#?}", error);
                             continue;
                         }
-                        let _control_value = control_value.unwrap();
+                        let control_value = control_value.unwrap();
 
-                        //TODO: Control V4L
+                        let mut param_result = mavlink::common::ParamAck::PARAM_ACK_ACCEPTED;
+                        if let Err(error) = mavlink_camera_information
+                            .as_ref()
+                            .lock()
+                            .unwrap()
+                            .video_source_type
+                            .inner()
+                            .set_control_by_id(control_id, control_value)
+                        {
+                            error!(
+                                "Failed to set parameter {control_id:?} with value {control_value:?}. Reason: {error:#?}",
+                            );
+                            param_result = mavlink::common::ParamAck::PARAM_ACK_FAILED;
+                        }
+
+                        if let Err(error) = vehicle.send(
+                            &header,
+                            &mavlink::common::MavMessage::PARAM_EXT_ACK(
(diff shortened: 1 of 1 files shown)


joaoantoniocardoso:
> Despite the errors on build, which are not happening on my machine, the ack answer isn't right.

Both fixed.

src/mavlink/mavlink_camera.rs
```diff
@@ -373,15 +373,41 @@ fn receive_message_loop(
...
+                            .set_control_by_id(control_id, control_value)
+                        {
+                            error!(
+                                "Failed to set parameter {:?} with value {:?}. Reason: {:#?}",
+                                control_id, control_value, error
+                            );
+                            param_result = mavlink::common::ParamAck::PARAM_ACK_FAILED;
+                        }
+
+                        if let Err(error) = vehicle.send(
```
joaoantoniocardoso, line 398:
```suggestion
                        }
                        drop(information);

                        if let Err(error) = vehicle.send(
```

src/mavlink/mavlink_camera.rs
```diff
@@ -373,15 +373,41 @@ fn receive_message_loop(
...
+                            .video_source_type
+                            .inner()
+                            .set_control_by_id(control_id, control_value)
+                        {
+                            error!(
+                                "Failed to set parameter {:?} with value {:?}. Reason: {:#?}",
+                                control_id, control_value, error
+                            );
+                            param_result = mavlink::common::ParamAck::PARAM_ACK_FAILED;
+                        }
```
patrickelectric, line 396: To avoid missing the drop, you can avoid creating an intermediary component since information is only used once in this scope. 
```suggestion
                        let mut param_result = mavlink::common::ParamAck::PARAM_ACK_ACCEPTED;
                        if let Err(error) = mavlink_camera_information
                            .as_ref()
                            .lock()
                            .unwrap()
                            .video_source_type
                            .inner()
                            .set_control_by_id(control_id, control_value)
                        {
                            error!(
                                "Failed to set parameter {:?} with value {:?}. Reason: {:#?}",
                                control_id, control_value, error
                            );
                            param_result = mavlink::common::ParamAck::PARAM_ACK_FAILED;
                        }
```

src/mavlink/mavlink_camera.rs
```diff
@@ -373,15 +373,41 @@ fn receive_message_loop(
...
-                        //TODO: Control V4L
+                        let information = mavlink_camera_information.as_ref().lock().unwrap();
+                        let mut param_result = mavlink::common::ParamAck::PARAM_ACK_ACCEPTED;
+                        if let Err(error) = information
+                            .video_source_type
+                            .inner()
+                            .set_control_by_id(control_id, control_value)
+                        {
+                            error!(
+                                "Failed to set parameter {:?} with value {:?}. Reason: {:#?}",
```
patrickelectric, line 392: you can use format args implicit here

patrickelectric, line 392: you can use format args implicit here

patrickelectric (changes requested):

PR #68, opened 2022-04-26
Fix unwanted access to CtrlClass

joaoantoniocardoso:
The following errors are caused by this unwanted access to CtrlClass:

image

### DIFF
--- a/src/video/video_source_local.rs
+++ b/src/video/video_source_local.rs
@@ -275,6 +275,14 @@ impl VideoSource for VideoSourceLocal {
                 },
                 ..Default::default()
             };
+
+            if matches!(v4l_control.typ, v4l::control::Type::CtrlClass) {
+                // CtrlClass is not a control, so we are skipping it to avoid any access to it, as it will raise an
+                // IO error #13: Permission Denied. To better understand, look for 'V4L2_CTRL_TYPE_CTRL_CLASS' on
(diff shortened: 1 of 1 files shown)


src/video/video_source_local.rs
```diff
@@ -275,6 +275,11 @@ impl VideoSource for VideoSourceLocal {
                 },
                 ..Default::default()
             };
+
+            if matches!(v4l_control.typ, v4l::control::Type::CtrlClass) {
+                continue; // Avoid trying to access a CtrlClass.
```
patrickelectric, line 280: Is there a why ? It would be nice to explain for people that is not aware, and a simple google search did not answer much.

joaoantoniocardoso: Yes, the reason is that CtrlClass is not a control, so it shouldn't be accessible by the same means. From [kernel.org](https://www.kernel.org/doc/html/v5.5/media/uapi/v4l/vidioc-queryctrl.html#c.v4l2_ctrl_type):

> This is not a control. When VIDIOC_QUERYCTRL is called with a control ID equal to a control class code (see [Control classes](https://www.kernel.org/doc/html/v5.5/media/uapi/v4l/vidioc-g-ext-ctrls.html#ctrl-class)) + 1, the ioctl returns the name of the control class and this control type. Older drivers which do not support this feature return an EINVAL error code. 
 
So it may be the rust library giving us something unnecessary.

patrickelectric (approved):

PR #77, opened 2022-05-20
Add UYUV and MPEG video encodings for all supported video sources and endpoints.

joaoantoniocardoso:
- [x] Manually tested on PC
- [x] Manually tested on Raspberry Pi


src/stream/stream_backend.rs
```diff
@@ -192,24 +187,29 @@ fn create_udp_stream(
...
-            } else {
-                return Err(SimpleError::new(format!(
-                    "Unsupported encode for UDP endpoint: {:?}",
-                    encode
-                )));
-            }
+            let capability = match &encode {
+                VideoEncodeType::H264 => "video/x-h264",
+                VideoEncodeType::YUYV => "video/x-raw,format=YUY2",
+                VideoEncodeType::MJPG => "image/jpeg",
```
patrickelectric, line 193: I just want to say that is beautiful

src/stream/stream_backend.rs
```diff
@@ -230,45 +230,51 @@ fn create_udp_stream(
...
+    let video_payload = match &encode {
+        VideoEncodeType::H264 => concat!(
             " ! h264parse",
             " ! queue",
             " ! rtph264pay config-interval=10 pt=96",
-        );
+        ),
+        VideoEncodeType::YUYV => concat!(
+            " ! videoconvert",
+            " ! video/x-raw,format=UYVY",
```
patrickelectric, line 252: this is already the format on `video_format`, can you explain a bit more about this here ? can you also add a comment ? This looks more like an issue since you are converting from the same input as the output.

src/stream/stream_backend.rs
```diff
@@ -316,24 +322,29 @@ fn create_rtsp_stream(
...
-            } else {
-                return Err(SimpleError::new(format!(
-                    "Unsupported encode for RTSP endpoint: {:?}",
-                    encode
-                )));
-            }
+            let capability = match &encode {
+                VideoEncodeType::H264 => "video/x-h264",
+                VideoEncodeType::YUYV => "video/x-raw,format=YUY2",
+                VideoEncodeType::MJPG => "image/jpeg",
```
patrickelectric, line 328: We should put that on a function that transform type to gstreamer type, this will remove the redundant text.

patrickelectric (changes requested):

src/stream/stream_backend.rs
joaoantoniocardoso: Here we need to transcode from _YUYV_ (_YUY2_ on gstreamer), which is the raw video format from the source, to _UYVY_, which is one of the raw format accepted by the payloader _rtpvrawpay_, since _YUY2_ isn't.

```bash
~ gst-inspect-1.0 rtpvrawpay | grep "format:"
                 format: { (string)RGB, (string)RGBA, (string)BGR, (string)BGRA, (string)AYUV, (string)UYVY, (string)I420, (string)Y41B, (string)UYVP }
```

Should it be commented on in the code?

patrickelectric: It would be nice to document that since it's a bit hard to identify that reading the code.

src/stream/stream_backend.rs
```diff
@@ -123,10 +123,6 @@ fn check_scheme(
                 }
             }
             "udp" => {
-                if VideoEncodeType::H264 != encode {
```
patrickelectric, line 126: There is no explanation on the commit body or in the title explaining why this is not necessary.

patrickelectric (changes requested):

src/stream/gst/pipeline_builder.rs
```diff
@@ -0,0 +1,225 @@
+use crate::{
```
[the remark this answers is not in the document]
joaoantoniocardoso: Yeah, I feel the same. This definitely accentuates the need for testing here, but it is as fragile as before - almost the only thing we are testing is if the constructed pipeline string is the same as expected. For sure it helps to some extent, but it is not enough.

I'll take a look at how pipelines are supposed to be tested, do you have any guidelines?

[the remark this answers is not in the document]
joaoantoniocardoso: I found two main strategies:
- 1. Intercept a test signal run through the pipeline, then compare it with the expected/reference, or
- 2. Generate output logs for the pipeline and compare them with the reference logs.

https://blogs.igalia.com/aboya/2019/05/14/validateflow-a-new-tool-to-test-gstreamer-pipelines/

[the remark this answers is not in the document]
patrickelectric: We can discuss the approaches later, I believe the current testing is enough for now.

src/stream/gst/pipeline_builder.rs
```diff
@@ -0,0 +1,225 @@
...
+#[derive(Clone, Debug, Default)]
+pub struct Pipeline {
+    pub description: String,
+}
+
+impl Pipeline {
+    pub fn new(
+        video_and_stream_information: &VideoAndStreamInformation,
+        configuration: &VideoCaptureConfiguration,
+        is_webrtcsink: bool,
```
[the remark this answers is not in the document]
joaoantoniocardoso: We can, but it would mean this _check if webrtc using endpoint_ code would be run multiple times, instead of only once.

src/stream/gst/pipeline_builder.rs
```diff
@@ -0,0 +1,225 @@
...
+        let video_source = &video_and_stream_information.video_source;
+
+        let capability = Pipeline::build_capability_string(video_source, &encode, &configuration)?;
+
+        let source = Pipeline::build_pipeline_source(video_source, &capability)?;
+        let transcode = Pipeline::build_pipeline_transcode(video_source, &encode, is_webrtcsink)?;
+        let payload = Pipeline::build_pipeline_payload(&encode, is_webrtcsink)?;
+        let sink = Pipeline::build_pipeline_sink(&endpoints, is_webrtcsink)?;
+
+        let description = format!("{source}{transcode}{payload}{sink}");
```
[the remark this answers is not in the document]
joaoantoniocardoso: Actually, I think this would be a good point to switch to using the GST API to "make" the elements, and then "link" them, [as they recommend doing](https://gstreamer.freedesktop.org/documentation/application-development/basics/elements.html?gi-language=c#linking-elements) instead of creating the entire pipeline as a string.

src/stream/gst/pipeline_builder.rs
```diff
@@ -0,0 +1,263 @@
...
+    fn build_pipeline_sink(
+        endpoints: &Vec<url::Url>,
+        is_webrtcsink: bool,
+        webrtc_name: Option<&String>,
+    ) -> Result<String, SimpleError> {
+        if is_webrtcsink {
+            let (stun_endpoint, turn_endpoint, signalling_endpoint) =
+                Pipeline::build_webrtc_endpoints(&endpoints)?;
+            let capability = "video/x-h264"; // We could also choose for video/x-vp9 here.
+            let webrtc_name = webrtc_name.unwrap();
```
patrickelectric, line 203: Why is this an optional ? Maybe because this is controller also by `is_webrtcsink` ? If so, there are better ways to handle it, like creating another function where the is_webrtcsink logic can exist.

jaxxzer:
so stoked good job @joaoantoniocardoso and @patrickelectric !

PR #83, opened 2022-06-10
Add h265 encoding

joaoantoniocardoso:
- [ ] Tested on Raspberry PI with a real camera.


joaoantoniocardoso:
Superseded by #252, although this PR also has TCP stuff which could be useful in the future.

Issue #66, opened 2022-04-20
Add support for raspberry camera


patrickelectric:
original comment: https://github.com/bluerobotics/BlueOS-docker/issues/991

joaoantoniocardoso:
Can we identify the raspberry pi camera with the current implementation?

ES-Alexander:
May be relevant:
https://www.raspberrypi.com/documentation/accessories/camera.html?ref=hackernoon.com#network-streaming

ES-Alexander:
> Can we identify the raspberry pi camera with the current implementation?

@joaoantoniocardoso `libcamera-hello --list-cameras` seems to detect mine fine, if that's helpful:
```
pi@blueos:~ $ libcamera-hello --list-cameras
[0:08:40.093741943] [1781]  INFO Camera camera manager.cpp:293 libcamera v0.0.0+3384-44d59841
[0:08:40.168778494] [1782]  WARN PI raspberrypi.cpp: 1202 Mismatch between Unicam and CamHelper for embedded data usage!
[0:08:40.169876612] [1782]  INFO PI raspberrypi.cpp: 1313 Registered camera /base/soc/i2c0mux/i2c@1/imx219@10 to Unicam device /dev/media3 and ISP device /dev/media0
Available cameras
----------------
0 : imx219 [3280x2464] (/base/soc/12c0mux/i2c@1/imx219@10)
    Modes: 'SBGGR10_CSI2P'  : 640x480 1640x1232 1920x1080 3280x2464
           'SBGGR8'  : 640x480 1640x1232 1920x1080 3280x2464
1 : H264 USB Camera: USB Camera [1920x1080] (/base/soc/usb@7e980000/usb1@1-1.2:1.0-32e4:9422)
    Modes: 'MJPEG'  : 320x240 352x288 640x360 640x480 800x600 1280x720 1920x1080 1920x1080
           'YUYV'  : 320x240 352x288 640x360 640x480 640×480 800x600
```

That said, that doesn't seem to detect the H264 output option, and I'm not certain how to ensure it does proper (efficient) hardware encoding like it does with earlier RPiOS versions.

PR #84, opened 2022-06-17
Add stepwise support

joaoantoniocardoso:
- [x] Tested on Raspberry Pi 3

This PR is one step further to #66. Actually, with this implementation, the raspberry pi camera works in legacy mode, but more test is needed.

### DIFF
--- a/src/video/video_source_local.rs
+++ b/src/video/video_source_local.rs
@@ -1,3 +1,5 @@
+use std::cmp::max;
+
 use super::types::*;
 use super::{
     video_source,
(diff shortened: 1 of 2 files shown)


src/video/video_source_local.rs
```diff
@@ -159,48 +182,87 @@ impl VideoSource for VideoSourceLocal {
...
+                            (1280, 1080),
+                            (1600, 1200),
+                            (1280, 720),
+                            (1024, 768),
+                            (960, 720),
+                            (800, 600),
+                            (640, 480),
+                            (640, 360),
+                            (320, 240),
+                            (256, 144),
```
joaoantoniocardoso, line 224: I've chosen to only add common and widely used resolutions as opposed to deciding resolutions algorithmically because I found it to be near impossible to balance the number of sizes with how useful they would actually be, every resolution looks random, like `1326x1768` for 4:3, and if we would be adding the standard ones, why should we even bother trying that...

src/video/video_source_local.rs
```diff
@@ -159,48 +182,87 @@ impl VideoSource for VideoSourceLocal {
...
                     }
+                    v4l::framesize::FrameSizeEnum::Stepwise(v4l_size) => {
+                        let std_sizes: Vec<(u32, u32)> = vec![
+                            (v4l_size.max_width, v4l_size.max_height),
+                            (7200, 3060),
+                            (7680, 4320),
+                            (3840, 2160),
+                            (2560, 1440),
+                            (1440, 1080),
+                            (1920, 1080),
```
patrickelectric, line 214: Needs sort, dedup and move this to a const.

patrickelectric (changes requested):

patrickelectric:
some comments would also be nice.

patrickelectric (changes requested):

src/video/video_source_local.rs
```diff
@@ -443,7 +443,11 @@ impl VideoSourceAvailable for VideoSourceLocal {
                 device_path: camera_path.clone(),
                 typ: VideoSourceLocalType::from_str(&caps.bus),
             };
-            cameras.push(VideoSourceType::Local(source));
+            if matches!(source.typ, VideoSourceLocalType::Unknown(_)) {
+                warn!("Ignoring camera with unknown Local Source Type: {source:#?}");
+            } else {
+                cameras.push(VideoSourceType::Local(source));
+            }
```
joaoantoniocardoso, line 450: @patrickelectric With this filter, we will have only the supported cameras available to the front-end, do you think it is better or worse to filter them here?

All those **unusable** devices like /dev/video15 and /dev/video23 that are created with the Raspberry Pi camera (e.g. /dev/video0) will be filtered out.

patrickelectric: We don't need to filter since it'll appears as unknown on the REST API. 
The API should provide everything that it can's and let the decision be taken on the frontend.

patrickelectric (approved):

PR #86, opened 2022-07-15
Add rpicamera support

### LINKED ISSUE #66: Add support for raspberry camera

joaoantoniocardoso:
Closes #66.

Finally, with this PR, Raspberry Pi Camera will be working, but with the following notes:
- Only Legacy Camera mode is supported, to do so, one needs to enable it using `raspi-config`, which can be done in one line: ` if [[ $(raspi-config nonint get_legacy) == 1 ]]; then sudo raspi-config nonint do_legacy 0 && sudo sync && sudo reboot; fi `
- Any Raspberry Pi Camera will be constrained to work with a maximum resolution of 1920 x 1080 px, and/or with a maximum frame rate of 30 FPS.
- rpicam needs to be added into the GStreamer build to have the `rpicamsrc` plugin available.

### DIFF
--- a/src/stream/gst/pipeline_builder.rs
+++ b/src/stream/gst/pipeline_builder.rs
@@ -88,7 +88,24 @@ impl Pipeline {
                 }
             },
             VideoSourceType::Local(local_device) => {
-                format!("v4l2src device={}", &local_device.device_path)
+                match &local_device.typ {
+                    crate::video::video_source_local::VideoSourceLocalType::Usb(_) => {
+                        format!("v4l2src device={}", &local_device.device_path)
+                    }
+                    // For raspberry pi cameras (in Legacy Mode), rpicamsrc
+                    // is used to be used insted v4l2src as a workaround to
+                    // prevent Kernel Oops caused by setting a unsupported but
+                    // wrongly stated as supported by v4l2 Frame Size and/or
+                    // FPS.
+                    crate::video::video_source_local::VideoSourceLocalType::LegacyRpiCam(_) => {
+                        format!("rpicamsrc camera-number=0")
+                    }
+                    typ => {
+                        return Err(SimpleError::new(format!(
+                            "Unsuported VideoSourceLocal: {typ:#?}."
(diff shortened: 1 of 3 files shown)


src/stream/gst/pipeline_builder.rs
```diff
@@ -88,7 +88,24 @@ impl Pipeline {
...
+                    // wrongly stated as supported by v4l2 Frame Size and/or
+                    // FPS.
+                    crate::video::video_source_local::VideoSourceLocalType::LegacyRpiCam(_) => {
+                        format!("rpicamsrc camera-number=0")
+                    }
+                    typ => {
+                        return Err(SimpleError::new(format!(
+                            "Unsuported VideoSourceLocal: {typ:#?}."
+                        )))
+                    }
```
joaoantoniocardoso, line 107: note that this changes the current logic, but all current supported cameras still work (USB cameras)

patrickelectric: Can you explain a bit more how this changes the logic ? Can you also put that in your commit body ?

joaoantoniocardoso: Sure. Originally, any `VideoSourceLocalType` variant - including `Usb` and `Unknown` - would result in a pipeline using `v4l2src`. Now, only the `Usb` variant will, meaning that `Unknown` is not supported. The idea here is to prevent scenarios like when using v4l2src with the Raspberry Pi camera in Legacy Camera mode, which easily causes Kernel Oops/Panic.

note that I intentionally left the tests untouched so we are obligated to have this discussion - the tests are creating a `crate::video::video_source_local::VideoSourceLocalType::Unknown("Potato")` camera, which will not be supported if we choose this change in behavior.

joaoantoniocardoso: @patrickelectric  ping

patrickelectric (approved):

PR #90, opened 2022-07-18
Make mavlink_endpoint optional 

### LINKED ISSUE #76: Add option to enable/disable mavlink
https://github.com/bluerobotics/BlueOS-docker/issues/1010

patrickelectric:
Fix #76


joaoantoniocardoso (approved):
> Fix #76

Does *partially* fix #76, as said in [here](https://github.com/bluerobotics/BlueOS-docker/issues/1010#issue-1233332360):

> In order of preference
> 1. A "Disable MAVLink" option _per endpoint_ of each camera
>    - neat/intuitive UI may be tricky, but this is the best for functionality
>    - preferably something like a checkbox beside each endpoint, but effective labelling may be tricky
>    - if necessary could even be added in the "Extra configuration" dropdown as an input for a comma-separated list of indices
> 2. A "Disable MAVLink" checkbox _per camera_
>    - e.g. in the existing "Extra configuration" dropdown
> 4. A "Disable MAVLink" checkbox for the entire camera manager
>    - seems much less versatile than the alternatives, but would at least solve the use-case in the forum post

So we can mark no.4 as done, but keep the rest.

PR #95, opened 2022-07-19
src: stream: Make the old settings compatible.

### LINKED ISSUE #81: Failing to load settings between versions
for more information: https://github.com/bluerobotics/BlueOS-docker/issues/1030

joaoantoniocardoso:
Turns out that we can make the tag "type" as optional for VideoCaptureConfiguration ("type": "video"), but required for any other, like RedirectCaptureConfiguration ("type": "redirect").

Fixes #81.

### DIFF
--- a/src/stream/types.rs
+++ b/src/stream/types.rs
@@ -50,10 +50,11 @@ pub struct VideoCaptureConfiguration {
 }
 
 #[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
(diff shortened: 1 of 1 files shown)


joaoantoniocardoso:
Draft: I'm still testing all cases.

patrickelectric:
That's a requirement to make it easy for the frontend to detect the object type inside the json. enums must be `tagged`

PR #102, opened 2022-07-26
cargo, src: video: Upgrade to v4l 0.13.0

joaoantoniocardoso:
- v4l 0.12.0 -> 0.13.0

### DIFF
--- a/src/video/video_source_local.rs
+++ b/src/video/video_source_local.rs
@@ -321,10 +321,10 @@ impl VideoSource for VideoSourceLocal {
         //TODO: Add control validation
         let device = Device::with_path(&self.device_path)?;
         //TODO: we should handle value, value64 and string
-        match device.set_control(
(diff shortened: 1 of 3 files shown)


patrickelectric (approved):
Good thing that we don't rely on v4l structs for the rest api

src/video/video_source_local.rs
```diff
@@ -385,7 +384,10 @@ impl VideoSource for VideoSourceLocal {
                 continue;
             }
             let value = value.unwrap();
-            let default = v4l_control.default;
+            let default = v4l_control.default.try_into().unwrap();
```
patrickelectric, line 387: `default` is an experimental keyword in rust, please use something else.

patrickelectric (changes requested):

patrickelectric (approved):

PR #97, opened 2022-07-26
src: video: Remove v4l2 controls for LegacyRpiCam.

joaoantoniocardoso:
v4l2 controls don't take effect on `rpicamsrc`, instead, the `rpicamsrc` element has several properties which can (probably) be changed in run time via GStreamer API. Until it is implemented here, there is no reason to expose any control for LegacyRpiCam local devices.

Fixes #96 

- [x] tested

PR #104, opened 2022-07-27
src: stream: Use v4l2src for LegacyRpiCam.

### LINKED ISSUE #96: Controls don't work for LegacyRpiCam
v4l2 controls don't take effect on `rpicamsrc` streams.

joaoantoniocardoso:
fixes #96, go against #97.

Instead of removing v4l2 controls because they don't work with `rpicamsrc` (PR #97), here we opt to use `v4l2src` instead of `rpicamsrc`, and keep the v4l2 controls working.

Why we were using `rpicamsrc` in the first place? In previous tests, `rpicamsrc` seems to be more stable than `v4l2src` in terms of avoiding 'kernel oops', but I gave `v4l2src` another chance and it seems to be as stable as `rpicamsrc`.

### DIFF
--- a/src/stream/gst/pipeline_builder.rs
+++ b/src/stream/gst/pipeline_builder.rs
@@ -3,6 +3,7 @@ use crate::{
     video::{
         types::{VideoEncodeType, VideoSourceType},
         video_source_gst::VideoSourceGstType,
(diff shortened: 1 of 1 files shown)


patrickelectric (commented):
nice!

Issue #114, opened 2022-08-17
RTSP is not working with QGC


joaoantoniocardoso:
The problem is very straightforward:

- We are now advertising via MAVLink the same IP as the user chooses as the stream endpoint when creating the stream
- We are still restricting RTSP streams to have "0.0.0.0" as the stream IP

As consequence, the user will never be able to have the QGC connect to the RTSP stream, only if using an additional redirect stream advertising the _adapter IP_.

Overall this IP thing was quite confusing to me at the beginning, and it might be as well for the end user: while UDP **only** works using the _network IP_ (like `192.168.2.1`), RTSP **only** works when using the _adapter IP_ (like `192.168.2.2`) - will the users know what to choose?

I see two ways to solve this and keep consistency:
A. Just allow the user to choose any IP for RTSP streams, which is what we are doing for any other stream, which isn't a big deal until now.
B. Choose the correct IP for each stream kind, instead of asking for the user to be an expert.

Today the user has to enter the correct IP for each stream kind he is setting up otherwise the stream will never work. Is there any specific reason (that I'm not seeing) for us to ask for the right IP for the users, instead of programmatically choosing the right one for them?

joaoantoniocardoso:
Right now, we are choosing to leave the responsibility for the front-end - the front-end should implement their own logic to solve this confusing setup. The BlueOs has implemented a good example (see https://github.com/bluerobotics/BlueOS-docker/pull/1125).

Since #115 this issue is not a bug anymore, so I'm closing it.

patrickelectric:
We can provide a mdns for QGC find the IP

PR #115, opened 2022-08-17
src: stream: Allow any IP for RTSP streams.

joaoantoniocardoso:
Now that we are advertising via MAVLink with the same IP as the stream endpoint, we should allow the use of the local IP for RTSP (like `192.168.2.2`) instead of restricting it to `0.0.0.0`, so QGC can connect to it using the advertised IP, otherwise, QGC would be trying to connect to `0.0.0.0`, which will fail if the RTSP server and QGC are not both under the same localhost.

This solves the bug related to #114, but that issue should be kept open for further discussion over automatic IP.

### DIFF
--- a/src/stream/stream_backend.rs
+++ b/src/stream/stream_backend.rs
@@ -173,12 +173,6 @@ fn create_rtsp_stream(
             endpoint.scheme()
         )));
     }
-    if endpoint.host_str() != "0.0.0.0".into() {
-        return Err(simple_error!(format!(
-            "The URL's host for RTSP endpoints should be \"0.0.0.0\", but was: {:?}",
-            endpoint.host_str()
-        )));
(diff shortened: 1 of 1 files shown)


patrickelectric (commented):
I'm not sure how this fix the issue, what is the use case of `0.0.0.0` ?

joaoantoniocardoso:
> I'm not sure how this fix the issue, what is the use case of `0.0.0.0` ?

The only use-case for 0.0.0.0 would be if the rtsp client is in the same computer as the rtsp server, which will not happen in BlueOs. Before we were requiring it to be "0.0.0.0" but we were advertising "192.168.2.2" via MAVLink, that's why it wasn't a concern.

patrickelectric (approved):

PR #116, opened 2022-08-17
src: mavlink: Don't try to restart connection on WouldBlock error kind.

joaoantoniocardoso:


src/mavlink/mavlink_camera.rs
```diff
@@ -321,6 +321,12 @@ fn heartbeat_loop(
         }
 
         if let Err(error) = vehicle.read().unwrap().send(&header, &heartbeat_message()) {
+            {
+                let mavlink::error::MessageWriteError::Io(io_error) = &error;
+                if io_error.kind() == std::io::ErrorKind::WouldBlock {
+                    continue;
+                }
+            }
```
patrickelectric, line 329: Why is this in is own scope and why it's not after the error log message ?

joaoantoniocardoso:
Closing because I'm not able to reproduce the WouldBlock in the Raspberry Pi running TCP with `mavlink-routerd`, it seems to be happening only in my testing setup with mavp2p. :facepalm: 

All related problems I was having with TCP in Raspberry Pi were due to a modification of this branch, not the current main branch.

PR #117, opened 2022-08-19
stream: manager: Do not crete streams if vector is empty

patrickelectric:
Only do that if the settings does not exist, or the user force the reset

Signed-off-by: Patrick José Pereira <patrickelectric@gmail.com>

### DIFF
--- a/src/settings/manager.rs
+++ b/src/settings/manager.rs
@@ -6,6 +6,7 @@ use std::path::Path;
 use std::sync::{Arc, Mutex};
 
 use crate::cli;
+use crate::custom;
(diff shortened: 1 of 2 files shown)


joaoantoniocardoso (changes requested):
This works, but as a side effect, no streams are being created when we have no settings file, for example, using `--default-settings BlueROVUDP`.

patrickelectric:
@joaoantoniocardoso Sorry, I forgot to add the default streams in settings.

joaoantoniocardoso (approved):
Perfect, thanks!

PR #121, opened 2022-08-29
Add log to file

joaoantoniocardoso:
- [x] Log to file:
    * Always on.
    * The path is configurable via CLI argument `--log-path`, default: `./logs/`.
    * The level for log is not configurable, always `debug`.
    * Files are rotated each day.
- [x] Log to console:
    * Always on.
    * Colored output.
    * With a filter configurable via `RUST_LOG` env or CLI argument `--verbose`.
    * The default level for console is `info`.
- [x] Logs actix-web

## Behavioral changes:
- Log to file
- Format of console log messages differs from the previous: it has all the old fields, has gained some new ones, and now it is much easier to understand what is going on.

## For future:
- Tracing enables structured logging, we can use it to group events and make sense of which function call is related to each request, and so on.
- Tracing ecosystem supports several structured logging formats and solutions.
- Filters can be dynamically changed, so we can create an endpoint to set custom filters for the console log, or enable tracing for a temporary file.


src/cli/manager.rs
```diff
@@ -51,6 +51,15 @@ pub fn mavlink_connection_string() -> Option<&'static str> {
     return MANAGER.as_ref().clap_matches.value_of("mavlink");
 }
 
+pub(crate) fn log_path() -> String {
```
patrickelectric, line 54: Wait, what is the difference between `pub fn` and `pub(crate)` ?

src/cli/manager.rs
```diff
@@ -51,6 +51,15 @@ pub fn mavlink_connection_string() -> Option<&'static str> {
...
+    MANAGER
+        .as_ref()
+        .clap_matches
+        .value_of("log-path")
+        .expect("Clap arg \"log-path\" should always be \"Some(_)\" because of the default value.")
```
patrickelectric, line 59: Ok, the expect makes sense here, but that's kinda of unnecessary.
I would recommend for us to move to [clap derive](https://docs.rs/clap/latest/clap/) in another PR to avoid such poor consequences of the current architecture of clap.

src/logger/manager.rs
```diff
@@ -1,43 +1,52 @@
...
+use tracing_log::LogTracer;
+use tracing_subscriber::{fmt, layer::SubscriberExt, prelude::*, EnvFilter, Layer};
 
 // Start logger, should be done inside main
 pub fn init() {
-    if cli::manager::is_verbose() {
-        std::env::set_var("RUST_LOG", "debug");
-    }
+    // Redirect all logs from libs using "Log"
+    LogTracer::init_with_filter(tracing::log::LevelFilter::Trace).expect("Failed to set logger");
```
patrickelectric, line 10: Is that line necessary ? Are we not changing the filter on the next line ?

src/logger/manager.rs
```diff
@@ -1,43 +1,52 @@
...
+
+    // Configure the console log
+    let env_filter = EnvFilter::try_from_default_env().unwrap_or_else(|_| {
+        if cli::manager::is_verbose() {
+            EnvFilter::new("debug")
```
patrickelectric, line 15: Can't we use `LevelFilter` here directly ?

src/logger/manager.rs
```diff
@@ -1,43 +1,52 @@
...
+    let file_layer = fmt::Layer::new()
+        .with_writer(file_appender)
+        .with_ansi(false)
+        .with_file(true)
+        .with_line_number(true)
+        .with_span_events(fmt::format::FmtSpan::FULL)
+        .with_target(false)
+        .with_thread_ids(true)
+        .with_thread_names(true)
+        .with_filter(EnvFilter::new("debug"));
```
patrickelectric, line 43: `LevelFilter`, this will make more safer since is based in enum and not strings.

src/cli/manager.rs
joaoantoniocardoso: We also should also update to a more recent clap :+1:

src/logger/manager.rs
joaoantoniocardoso: Nope, here we are configuring the middleware to redirect logs from any library built using the log crate, this filter is different from the next ones, although Trace is not really filtering anything, just trying to be clearer about what level is being redirected (any level).

src/logger/manager.rs
joaoantoniocardoso: Unfortunately, they aren't the same thing. A filter uses levels, but isn't restricted to them, we could, for example, set `info` for our `mavlink` module, and trace for our `server` module, as `mavlink=debug,server=trace`.

src/logger/manager.rs
joaoantoniocardoso: see https://github.com/bluerobotics/mavlink-camera-manager/pull/121#discussion_r957380663

src/cli/manager.rs
joaoantoniocardoso: lol, I didn't see that, pub(crate) is public only for the crate, useful for libraries I guess, it was generated by the vscode, I'll change it.

src/logger/manager.rs
joaoantoniocardoso:
```suggestion
            EnvFilter::new(LevelFilter::DEBUG.to_string())
```
better?

src/logger/manager.rs
joaoantoniocardoso:
```suggestion
        .with_filter(LevelFilter::DEBUG);
```

patrickelectric (approved):

PR #123, opened 2022-08-29
src: mavlink: Remove 10ms delay from normal execution flow.

joaoantoniocardoso:


patrickelectric (approved):
Hope that does not end up in a spin lock cpu usage
merge at will

joaoantoniocardoso:
> Hope that does not end up in a spin lock cpu usage merge at will

It works just as before I've added in the TCP PR

Issue #14, opened 2021-12-29
About the duplicated device


joaoantoniocardoso:
What seems to be 'duplicated devices' on our frontend is actually somewhat an extended feature from [Linux Kernel](https://unix.stackexchange.com/a/539573). 

![image](https://user-images.githubusercontent.com/5920286/147662559-c9d3fe48-f5a2-4818-aaa6-495dccf658ec.png)

Because the user can't use (I'm assuming this based on my tests) both videos from an unique _bus_, and the commands from one affects the others, I think we should deal with it, and use it in a more concise way.

My initial thought on that is that maybe we could merge all the /dev/video* that shares the same _bus_, but probably we can't simply merge the controls/capabilities and expect it to works using the devices interchangeably - lets say video0 has JPEG encode, and to the video2 has H264, but if we merge both and keep only the video0 reference, can we use video0 as H264? Probably not, then we would need to add an abstraction layer just to map each capability to its original device (in this example, map JPEG to video0 and H264 to video2).

Maybe we could change our model to something that encapsulates each device inside each bus:
```
{
'': {
    'bus': 'usb-0000:09:00.3-2.4',
    'devices': [
      {
        'name': 'Video Capture 6',
        'device_path': '/dev/video2',
        ...
      },
      {
        'name': 'Video Capture 5',
        'device_path': '/dev/video0',
        ...
      }
    ]
    'controls': ...,
    'formats': ...,
    ...
  }
}
```

Do we want to deal with it, or we'll let it to the user? Any ideas?

patrickelectric:
Usually each device has different capabilities, as video stream types and etc.

PR #125, opened 2022-08-30
Merge devices from the same bus

joaoantoniocardoso:
- Helps #14: gives enough information for the front-end to merge the devices from the same BUS if it wants to.
- Also gives enough information for the front-end to detect each port a USB camera is connected to (v4l_by_path).

---
#### 1. `VideoSourceLocal` is used in the API:

image

#### Diff

```diff
#[derive(Apiv2Schema, Clone, Debug, Deserialize, PartialEq, Serialize)]
pub struct VideoSourceLocal {
    pub name: String,
    pub device_path: String,
+    pub v4l_by_path: String,
    #[serde(rename = "type")]
    pub typ: VideoSourceLocalType,
}
```

Because of this change on `VideoSourceLocal`, settings will also break.

That's all

### DIFF
--- a/src/server/pages.rs
+++ b/src/server/pages.rs
@@ -25,6 +25,7 @@ use std::io::prelude::*;
 pub struct ApiVideoSource {
     name: String,
     source: String,
+    source_type: VideoSourceType,
     formats: Vec<Format>,
     controls: Vec<Control>,
 }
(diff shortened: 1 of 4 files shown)


patrickelectric:
@joaoantoniocardoso can you specify what exactly this breaks ?

joaoantoniocardoso:
> @joaoantoniocardoso can you specify what exactly this breaks ?

[Done](https://github.com/bluerobotics/mavlink-camera-manager/pull/125#issue-1355658385)

patrickelectric:
Example of how to create the same desired functionality: 

https://github.com/patrickelectric/mavlink-camera-manager/pull/new/example-device

We can deal with the device in the frontend with:
```qml
property var device: devices
  .filter(device => device.real_device == box_name.currentValue)
  .map(device => {
     device.formats = device.formats.map(format => {
          format.source = device.source; return format
     })
     return device.formats
  })
```

PR #128, opened 2022-08-31
src: stream: Ensure StreamBackend also have Drop implemention.

joaoantoniocardoso:
I was studying when I saw this simple solution.

Solves #35

### DIFF
--- a/src/stream/stream_backend.rs
+++ b/src/stream/stream_backend.rs
@@ -8,7 +8,10 @@ use crate::video::types::{VideoEncodeType, VideoSourceType};
 use crate::video_stream::types::VideoAndStreamInformation;
 use simple_error::{simple_error, SimpleError};
 
-pub trait StreamBackend {
+pub trait StreamBackend
+where
+    Self: Drop,
+{
     fn start(&mut self) -> bool;
     fn stop(&mut self) -> bool;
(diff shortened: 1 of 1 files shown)


src/stream/stream_backend.rs
```diff
@@ -8,7 +8,10 @@ use crate::video::types::{VideoEncodeType, VideoSourceType};
 use crate::video_stream::types::VideoAndStreamInformation;
 use simple_error::{simple_error, SimpleError};
 
-pub trait StreamBackend {
+pub trait StreamBackend
+where
+    Self: Drop,
```
patrickelectric, line 13: What about the warning ?
```
bounds on `Self: std::ops::Drop` are most likely incorrect, consider instead using `std::mem::needs_drop` to detect whether a type can be trivially dropped
```

joaoantoniocardoso: From my understanding, it is just saying that we might don't need drop because the types would that implement StreamBackend trait might have trivial drop.

patrickelectric: Please, if possible, it would be nice to add `#[allow(drop_bounds)]`

patrickelectric (approved):

PR #132, opened 2022-09-07
Create ROV RTSP default settings

joaoantoniocardoso:
- [x] Tested on Raspberry Pi

This patch enhances our support for DHCP setups by:
1. Allowing the use of a Dynamic DNS, configured via CLI argument, to be used as the URI for the Mavlink camera XML file, as well in the RTSP stream URI, instead of directly using the local IP.
2. Allowing the creation of RTSP streams by default, opt-in via CLI argument.

---

Example:

image

### DIFF
--- a/src/custom/bluerov.rs
+++ b/src/custom/bluerov.rs
@@ -1,5 +1,6 @@
 use url::Url;
 
+use crate::network::utils::get_visible_qgc_address;
 use crate::stream::types::*;
 use crate::video::{self, types::*, video_source::VideoSourceAvailable};
 use crate::video_stream::types::VideoAndStreamInformation;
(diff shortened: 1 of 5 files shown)


src/custom/bluerov.rs
```diff
@@ -48,3 +50,57 @@ pub fn udp() -> Vec<VideoAndStreamInformation> {
...
+                    .last()
+                    .unwrap_or(&std::net::Ipv4Addr::UNSPECIFIED)
+                    .to_string(),
+            };
+
+            VideoAndStreamInformation {
+                name: format!("RTSP Stream {index}"),
+                stream_information: StreamInformation {
+                    endpoints: vec![Url::parse(&format!(
+                        "rtsp://{visible_qgc_ip_address}:8554/{index}"
```
patrickelectric, line 91: I would recommend to follow BlueOS format, that is `veideo_{index}`

patrickelectric (changes requested):

joaoantoniocardoso:
Just encapsulating the qgc ip into a function...

patrickelectric (approved):

PR #133, opened 2022-09-07
Reduce log size

joaoantoniocardoso:


src/server/manager.rs
```diff
@@ -49,7 +49,7 @@ pub async fn run(server_address: &str) -> Result<(), std::io::Error> {
 
         // Add debug call for API access
         app.wrap_fn(|req, srv| {
-            debug!("{:#?}", &req);
+            trace!("{:#?}", &req);
```
joaoantoniocardoso, line 52: All requests are still being displayed in INFO level, but with less information.

patrickelectric (approved):

PR #137, opened 2022-09-14
Fix settings

### LINKED ISSUE #81: Failing to load settings between versions

joaoantoniocardoso:
Closes #81 by resetting the settings if incompatible, but before, it creates a simple copy as a backup.

### DIFF
--- a/src/settings/manager.rs
+++ b/src/settings/manager.rs
@@ -110,14 +110,15 @@ fn fallback_settings_with_backup_file(file_name: &str) -> SettingsStruct {
 
 fn load_settings_from_file(file_name: &str) -> SettingsStruct {
     let result = std::fs::read_to_string(file_name);
-
-    debug!("loaded!");
-    if result.is_err() || cli::manager::is_reset() {
-        return SettingsStruct::default();
+    if let Err(error) = result {
+        error!("Error loading settings file {file_name:?}. Reason: {error:?}");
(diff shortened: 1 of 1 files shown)


src/settings/manager.rs
```diff
@@ -110,14 +110,15 @@ fn fallback_settings_with_backup_file(file_name: &str) -> SettingsStruct {
 
 fn load_settings_from_file(file_name: &str) -> SettingsStruct {
     let result = std::fs::read_to_string(file_name);
-
-    debug!("loaded!");
-    if result.is_err() || cli::manager::is_reset() {
```
patrickelectric, line 106: What about the reset ?

src/settings/manager.rs
```diff
@@ -72,13 +72,17 @@ impl Manager {
             file_name.into()
         };
 
-        debug!("Using settings file: {}", &file_name);
-
-        let settings = load_settings_from_file(&file_name);
+        let config = if cli::manager::is_reset() {
```
patrickelectric, line 75: I see

patrickelectric (approved):

src/settings/manager.rs
joaoantoniocardoso: It has moved to the function that calls it

PR #138, opened 2022-09-16
Update webrtc

joaoantoniocardoso:
note: this is incompatible with the current `libwebrtcsink.so` in BlueOS, an updated version must be uploaded, and both docker-base and blueos-docker need to be rebuilt.


patrickelectric:
This PR should be close since we talked that we are not going to use webrtcsink for now, and going to create our own solution to remove the camera encoding from our side.

PR #142, opened 2022-11-14
src, cargo: Completely remove WebRTCSink.

joaoantoniocardoso:
Removing WebRTC support until add our own implementation using WebRTCBin


src/server/manager.rs
```diff
@@ -20,31 +18,18 @@ fn json_error_handler(error: JsonPayloadError, _: &HttpRequest) -> actix_web::Er
...
-                info!("Using the www-path: {path:?}.");
-                app = app.service(
-                    actix_files::Files::new("webrtc", format!("{path}/webrtc/"))
-                        .index_file("index.html"),
-                );
-            }
+            // Load static files from configured www-path
+            info!("Using the www-path: {path:?}.");
+            app = app.service(
+                actix_files::Files::new("webrtc", format!("{path}/webrtc/"))
```
patrickelectric, line 30: Why is this necessary ?

joaoantoniocardoso: It's not necessary indeed

src/stream/mod.rs
```diff
@@ -6,5 +6,4 @@ pub mod types;
 pub mod video_stream_redirect;
 pub mod video_stream_rtsp;
 pub mod video_stream_udp;
-pub mod video_stream_webrtc;
 pub mod webrtc;
```
patrickelectric, line 9: why this was not removed ?

patrickelectric (changes requested):
It appears that multiple parts of the webrtc are still in the codebase with this PR.
That also includes turn and tokio, anyhow and util

src/stream/mod.rs
patrickelectric: And about anyhow ?

joaoantoniocardoso: I left anyhow because I am using it in our new WebRTC implementation, but I can remove and add again later if it makes more sense

patrickelectric: Ok, it's fine since it was explained. But usually removing and adding back is the cleaner solution, since we can't predict future code. 
We should also add a linter to notify about unused includes.

patrickelectric (approved):

rafaellehmkuhl:
Just to make sure, one using cockpit+webrtc shouldn't update from this point, till we merge the new implementation (in both camera manager and cockpit), right?

patrickelectric:
We're not going to update the camera manager on BlueOS before finishing the new webrtc implementation

PR #149, opened 2022-11-29
Fix CI

joaoantoniocardoso:
The entire CI is now taking: 
- ~10 minutes when not cached (~18x faster than before)
- ~5 minutes when cached (~12x faster than before)


joaoantoniocardoso:
Question: Why are we running the deploy CI twice (`on: [push, pull_request]`)?

Cross.toml
```diff
@@ -0,0 +1,5 @@
+[target.x86_64-unknown-linux-gnu]
+image = "joaoantoniocardoso/cross-rs:x86_64-unknown-linux-gnu-22.04-with-gstreamer"
```
joaoantoniocardoso, line 2: TODO: update the dockerhub link once we upload to bluerobotics' dockerhub

.github/workflows/deploy-unix.yml
```diff
@@ -6,35 +6,41 @@ jobs:
...
-      run: |
-        uname -a
-        sudo apt update
-        sudo apt install -y libunwind-dev
-        sudo apt install -y libclang-dev openssl
-        sudo apt install -y libgstrtspserver-1.0-dev libavahi-glib-dev libavahi-core-dev
+    - name: Use cached dependencies
+      uses: Swatinem/rust-cache@v2
+      with:
+        key: "${{ matrix.os }}-${{ matrix.TARGET }}-${{ hashFiles('**/Cargo.lock') }}"
```
joaoantoniocardoso, line 35:
```suggestion
        key: |
          "${{ matrix.os }}-${{ matrix.TARGET }}-${{ hashFiles('**/Cargo.lock') }}"
          "${{ matrix.os }}-${{ matrix.TARGET }}"
          "${{ matrix.os }}"
        
```

patrickelectric (approved):

.github/workflows/deploy-unix.yml
patrickelectric: Since we are unsure that the action may support multiple keys we can let as it stands for now and try it in the future with more time.

PR #155, opened 2022-12-13
Investigation: Add new RTSP

joaoantoniocardoso:


src/stream/stream.rs
```diff
@@ -37,102 +38,122 @@ impl Stream {
...
+        if endpoints.iter().any(|endpoint| endpoint.scheme() == "rtsp") {
+            if let Err(reason) =
+                create_rtsp_sink(Manager::generate_uuid(), video_and_stream_information)
+                    .and_then(|sink| stream.pipeline.add_sink(sink))
+            {
+                return Err(anyhow!(
+                    "Failed to add Sink of type RTSP to the Pipeline. Reason: {reason}"
+                ));
+            }
+        }
```
joaoantoniocardoso, line 68: The difference here is that we only want to add 1 UDP Sink (`multiudpsink`) even if we have several UDP endpoints. The same idea applies to RTSP, which, in the end only accepts 1 endpoint.

src/stream/rtsp/rtsp_server.rs
```diff
@@ -0,0 +1,186 @@
...
+        // test endpoint, providing the pipeline from the cli.
+        let mounts = rtsp_server
+            .server
+            .mount_points()
+            .context("Could not get mount points")?;
+
+        mounts.remove_factory(path);
+
+        rtsp_server.path_to_factory.remove(path);
+        // TODO: call mainloop.quit() to stop the server if there is no endpoints
```
patrickelectric, line 166: is that really necessary ?
can't we have it running ? It should use no cpu at all and the code will be simpler.

src/stream/stream.rs
```diff
@@ -37,102 +38,124 @@ impl Stream {
...
-    Ok(())
-}
+    if endpoints.iter().fold(0, |count, endpoint| {
+        if endpoint.scheme().eq("rtsp") {
+            count + 1
+        } else {
+            count
+        }
+    }) > 1
+    {
```
patrickelectric, line 89:
```suggestion
    if endpoints.iter().filter(|e| e.scheme() == "rtsp").count() > 1 {

```

patrickelectric (changes requested):
A quick review

src/stream/rtsp/rtsp_server.rs
joaoantoniocardoso: Indeed its not really necessary :)

Cargo.toml
```diff
@@ -57,6 +57,7 @@ tokio = { version = "1.21", features = ["full"] }
 enum_dispatch = "0.3.8"
 uuid = { version = "0.8", features = ["v4", "serde"] }
 ts-rs = "6.2"
+once_cell = "1"
```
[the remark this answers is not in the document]
joaoantoniocardoso: Nope, that was necessary when we were creating our custom factory :) thanks

src/stream/sink/rtsp_sink.rs
```diff
@@ -0,0 +1,165 @@
...
+            .property("flush-on-eos", true)
+            .property("max-size-buffers", 0u32) // Disable buffers
+            .build()?;
+
+        let path = addresses
+            .iter()
+            .find_map(|address| (address.scheme() == "rtsp").then_some(address.path().to_string()))
+            .context(
+                "Failed to find RTSP compatible address. Example: \"rtsp://0.0.0.0:8554/test\"",
+            )?;
```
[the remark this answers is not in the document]
joaoantoniocardoso: yeah, that's true, I think I'll use the stream id instead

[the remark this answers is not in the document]
joaoantoniocardoso: Now the `socket_path` lies inside the RtspSink struct, being constructed with the sink id, and I also remembered to move the the file removal to before/after the sink link and unlink.

patrickelectric (approved):

PR #159, opened 2022-12-17
src: sink: Implement SDP generation for sink interface

### LINKED ISSUE #17: Create sdp files for stream (Open stream on VLC)
This helps vlc and others programs to run streams without much user configuration

joaoantoniocardoso:
* UDP - works with VLC and GStreamer for H264, RAW, and MJPG.
* RTPS - both VLC and GStreamer parse it, but they refuse to play it.

closes #17

### DIFF
--- a/src/stream/sink/udp_sink.rs
+++ b/src/stream/sink/udp_sink.rs
@@ -14,6 +14,7 @@ pub struct UdpSink {
     udpsink: gst::Element,
     udpsink_sink_pad: gst::Pad,
     tee_src_pad: Option<gst::Pad>,
+    addresses: Vec<url::Url>,
 }
 impl SinkInterface for UdpSink {
     #[instrument(level = "debug")]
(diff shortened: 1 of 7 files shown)


patrickelectric (commented):
Can we add an endpoint that provide the sdp files ?

joaoantoniocardoso:
image

src/stream/sink/udp_sink.rs
```diff
@@ -142,7 +182,7 @@ impl UdpSink {
             })
             .collect::<Vec<String>>()
             .join(",");
-        let description = format!("multiudpsink clients={addresses}");
```
patrickelectric, line 145: try to avoid creating unnecessary noise in commits that are unrelated to the functional change.
you can do it in individual commits.

patrickelectric (approved):

src/stream/sink/udp_sink.rs
joaoantoniocardoso: This is related: because the property self.addresses was added, the meaning of `addresses` would be confusing in the context of this function.

patrickelectric: Oh, I see it now, thanks!

PR #168, opened 2023-01-12
src: stream: Filter out non-H264 local streams for WebRTC

### LINKED ISSUE #164: non-H264 (RAW and MJPG) streams should not be available for WebRTC

joaoantoniocardoso:
Closes #164

### DIFF
--- a/src/stream/webrtc/signalling_server.rs
+++ b/src/stream/webrtc/signalling_server.rs
@@ -300,19 +300,26 @@ impl StreamManagementInterface<Stream> for SignallingServer {
 
         streams
             .iter()
-            .map(|stream| {
+            .filter_map(|stream| {
                 let (height, width, encode, interval) =
                     match &stream.video_and_stream.stream_information.configuration {
-                        crate::stream::types::CaptureConfiguration::Video(configuration) => (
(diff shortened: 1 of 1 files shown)


src/stream/webrtc/signalling_server.rs
```diff
@@ -300,19 +300,25 @@ impl StreamManagementInterface<Stream> for SignallingServer {
...
-                            Some(
-                                (configuration.frame_interval.numerator as f32
-                                    / configuration.frame_interval.denominator as f32)
-                                    .to_string(),
-                            ),
-                        ),
+                        crate::stream::types::CaptureConfiguration::Video(configuration) => {
+                            // Filter out non-H264 local streams
+                            if configuration.encode != crate::video::types::VideoEncodeType::H264 {
+                                return None;
```
patrickelectric, line 309: We should also put a message saying that it was filtered.

joaoantoniocardoso: I can add a tracing level, otherwise, it will be shown for every webrtc's streams available request

patrickelectric (approved):

Issue #79, opened 2022-06-01
No errors when creating a stream with wrong Size/FPS


joaoantoniocardoso:
Related to #78.

When creating a stream using the wrongly associated Size/FPS for YUYV (eg. 1920x1080@30FPS), no errors are returned, and neither new stream populates the Stream list, but when trying to create another stream, it returns two possible errors:
- Stream name already in use
- Device already in use

Thus, indicating a bug on the back-end side.

joaoantoniocardoso:
#172 partially fixes the problem, but the error returned isn't good enough:
```
"Failed to add Sink of type UDP to the Pipeline. Reason: Failed setting Pipeline 283ec88f-3c8d-40a2-8ac8-2e10aaca8246 to Null state. Reason: set state timed-out (2 seconds)"
```

So we should add a validator to check camera capabilities before creating the v4l pipeline.

PR #171, opened 2023-01-13
src: html: Fix alerts not showing

joaoantoniocardoso:
Helps #79

### DIFF
--- a/src/html/index.html
+++ b/src/html/index.html
@@ -162,7 +162,11 @@
                 },
                 checkResponse: async function(response) {
                     if (!response.ok) {
-                        const text = await response.text().replaceAll("\\n", "\n").replaceAll("\\\"", "\"")
+                        // To make the alert text more human readable, here we are:
+                        //   1. removing the external double quotes pair, making the string `"text"` turns into `text`
+                        //   2. unescaping new lines, making the string `text\\n newline` turns into `text\n newline`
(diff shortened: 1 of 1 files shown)


src/html/index.html
```diff
@@ -162,7 +162,7 @@ <h3>Name: {{ stream.video_and_stream.name }}</h3>
                 },
                 checkResponse: async function(response) {
                     if (!response.ok) {
-                        const text = await response.text().replaceAll("\\n", "\n").replaceAll("\\\"", "\"")
+                        const text = await response.text().then((text) => text.replace(/^"(.+(?="$))"$/, '$1').replaceAll("\\n", "\n").replaceAll("\\\"", "\""))
```
patrickelectric, line 165: Can you be a bit more clear about what this regex is replacing ?

patrickelectric (changes requested):

src/html/index.html
joaoantoniocardoso: In sequence:
1. get content from inside the external double quotes pair
2. unescape new lines
3. unescape double quotes

patrickelectric: Ok, I would recommend to add a comment for the first regex and a minimal example of what it should match and what is the content of text just for the clarity. A look head/behind regex is never trivial.

joaoantoniocardoso: that's fair

joaoantoniocardoso: @patrickelectric done

patrickelectric (approved):

PR #186, opened 2023-02-21
cargo, src: Integrate GStreamer logs into Tracing

joaoantoniocardoso:
Without integrate_spans() because it would significantly change the format of the GST logs we are already used to, also difficulting web search.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2108,6 +2108,7 @@ dependencies = [
  "tracing",
  "tracing-actix-web",
  "tracing-appender",
+ "tracing-gstreamer",
  "tracing-log",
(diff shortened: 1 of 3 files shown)


patrickelectric:
@joaoantoniocardoso, just to make sure, we are still going to have gstreamer output on the terminal, right ?

joaoantoniocardoso:
> @joaoantoniocardoso, just to make sure, we are still going to have gstreamer output on the terminal, right ?

Yes, if the verbosity level doesn't filter it out: if we are looking for GST info messages, we need to enable the Rust info level.

patrickelectric (approved):

PR #193, opened 2023-03-15
src: custom: bluerov: Deal when there is no valid size available

patrickelectric:


src/custom/bluerov.rs
```diff
@@ -4,6 +4,7 @@ use crate::network::utils::get_visible_qgc_address;
 use crate::stream::types::*;
 use crate::video::{self, types::*, video_source::VideoSourceAvailable};
 use crate::video_stream::types::VideoAndStreamInformation;
+use log::*;
```
joaoantoniocardoso, line 7:
```suggestion
use tracing::*;
```

src/custom/bluerov.rs
```diff
@@ -28,9 +29,13 @@ pub fn udp() -> Vec<VideoAndStreamInformation> {
...
                     .cmp(&(10 * second_size.width + second_size.height))
             });
+            if sizes.last().is_none() {
+                warn!("Unable to find a valid size for {:?}", cam);
+                return None;
+            }
             let size = sizes.last().unwrap();
 
-            VideoAndStreamInformation {
+            Some(VideoAndStreamInformation {
```
joaoantoniocardoso, line 38: line 48:

```diff
---            }
+++            })
```

joaoantoniocardoso: And at line 19 and 68:
```diff
---        .map(|(index, cam)| {
+++        .filter_map(|(index, cam)| {
```

src/custom/bluerov.rs
```diff
@@ -93,7 +98,8 @@ pub fn rtsp() -> Vec<VideoAndStreamInformation> {
                     extended_configuration: None,
                 },
                 video_source: cam.clone(),
-            }
+            })
         })
+        .filter_map(|result| result)
```
joaoantoniocardoso, line 103:
```suggestion
```

joaoantoniocardoso (changes requested):
Instead of adding a filter_map at the end, replace the map with a map_filter in those two functions, also, there is some missing closing parenthesis.

joaoantoniocardoso (approved):
I feel that the first filter step is probably unnecessary, but this is not a critical section of the code anyways :)

src/custom/bluerov.rs
```diff
@@ -69,15 +81,17 @@ pub fn rtsp() -> Vec<VideoAndStreamInformation> {
...
-                (10 * first_size.width + first_size.height)
-                    .cmp(&(10 * second_size.width + second_size.height))
-            });
+            sort_sizes(&mut sizes);
+
+            if sizes.last().is_none() {
+                warn!("Unable to find a valid size for {:?}", cam);
+                return None;
+            }
             let size = sizes.last().unwrap();
```
joaoantoniocardoso, line 90: this could be:
```suggestion
            let Some(size) = sizes.last() else {
                warn!("Unable to find a valid size for {cam:?}");
                return None;
            };
```

PR #196, opened 2023-03-16
src: custom: bluerov: Deal when there is no valid size available

patrickelectric:


src/custom/bluerov.rs
```diff
@@ -44,7 +49,8 @@ pub fn udp() -> Vec<VideoAndStreamInformation> {
                     extended_configuration: None,
                 },
                 video_source: cam.clone(),
-            }
+            })
         })
+        .filter_map(|result| result)
```
joaoantoniocardoso, line 54: Remove this and replace the .map of line 18 with .filter_map

joaoantoniocardoso (approved):

PR #199, opened 2023-03-22
src: Make clippy happier

joaoantoniocardoso:


joaoantoniocardoso:
> Why are we not running it on CI ?

We are :eyes:

PR #201, opened 2023-03-22
Dont panic

joaoantoniocardoso:


joaoantoniocardoso:
Drafting 'cause I need to change some of those strings.

src/mavlink/mavlink_camera.rs
```diff
@@ -695,7 +767,13 @@ fn receive_message_loop(
...
                         }
 
-                        let information = mavlink_camera_information.lock().unwrap();
+                        let information = match mavlink_camera_information.lock() {
+                            Ok(guard) => guard,
+                            Err(error) => {
+                                error!("Failed locking a Mutex. Reason: {error}");
+                                return;
+                            }
+                        };
```
patrickelectric, line 776: We should create a macro for this

patrickelectric (approved):

src/mavlink/mavlink_camera.rs
joaoantoniocardoso: That's true. I should start writing small macros like that

PR #216, opened 2023-04-20
src: stream: Nicely remove all streams before create the defaults

### LINKED ISSUE #215: Pipelines are not being correctly killed when calling reset

joaoantoniocardoso:
Fixes #215

### DIFF
--- a/src/stream/manager.rs
+++ b/src/stream/manager.rs
@@ -87,14 +87,36 @@ fn config_gst_plugins() {
     }
 }
 
+pub fn remove_all_streams() -> Result<()> {
+    let keys = {
+        let mut ids = vec![];
+
+        match MANAGER.lock() {
+            Ok(guard) => guard,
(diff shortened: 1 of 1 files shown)


src/stream/manager.rs
```diff
@@ -87,14 +87,45 @@ fn config_gst_plugins() {
...
+    // First of all, gently remove all streams as we are going to replace the entire list below
+    remove_all_streams()?;
+
+    {
+        match MANAGER.lock() {
+            Ok(guard) => guard,
+            Err(error) => return Err(anyhow!("Failed locking a Mutex. Reason: {error}")),
+        }
+        .streams
+        .clear();
```
patrickelectric, line 127: why remove_all_streams is not supposed to do this ?

joaoantoniocardoso: who'd left this here? :eyes:

patrickelectric (approved):

PR #225, opened 2023-04-28
src: Make clippy happier

joaoantoniocardoso:


joaoantoniocardoso:
lesson: I should run clippy way often

patrickelectric:
> lesson: I should run clippy way often

Why are we not running on CI ?

joaoantoniocardoso:
> > lesson: I should run clippy way often
> 
> Why are we not running on CI ?

It is, but it lacks permission to make annotations

patrickelectric:
Did you check this ?
https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

patrickelectric (approved):

PR #233, opened 2023-05-24
src: stream: pipeline: Don't pause redirect's rtspsrc.

### LINKED ISSUE #229: Fail when recreating a WebRTC stream from some redirect sources
Fail when recreating a WebRTC stream from some redirect sources. It's happening with IP (some?) cameras, but not with RTSP sent by GStreamer, and it doesn't *always* happen.

Steps to reproduce:

1. Create a redirected stream from an IP camera
2. Connect via WebRTC
3. Disconnect
4. Try to connect again

joaoantoniocardoso:
Closes #229 

This is a workaround as it's better to keep consuming resources than have to restart the whole stream just because the client closed the stream.

### DIFF
--- a/src/stream/pipeline/mod.rs
+++ b/src/stream/pipeline/mod.rs
@@ -238,15 +238,24 @@ impl PipelineState {
         sink.unlink(pipeline, pipeline_id)?;
 
         // Set pipeline state to NULL when there are no consumers to save CPU usage.
-        let sink_name = format!("{PIPELINE_SINK_TEE_NAME}-{pipeline_id}");
-        let tee = pipeline
-            .by_name(&sink_name)
-            .context(format!("no element named {sink_name:#?}"))?;
-        if tee.src_pads().is_empty() {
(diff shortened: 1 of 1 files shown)


patrickelectric (approved):

src/stream/pipeline/mod.rs
```diff
@@ -238,15 +238,24 @@ impl PipelineState {
...
-        let tee = pipeline
-            .by_name(&sink_name)
-            .context(format!("no element named {sink_name:#?}"))?;
-        if tee.src_pads().is_empty() {
-            if let Err(error) = pipeline.set_state(gst::State::Null) {
-                return Err(anyhow!(
-                    "Failed to change state of Pipeline {pipeline_id} to NULL. Reason: {error}"
-                ));
+        // TODO: We are skipping rtspsrc here because once back to null, we are having
+        // trouble knowing how to propper reuse it.
```
joaoantoniocardoso, line 242: We should address this in a new issue

PR #239, opened 2023-06-09
src: mavlink: Avoid Clone to avoid unwanted Drops

### LINKED ISSUE #238: Advertising wrong MAVLink ID
image

joaoantoniocardoso:
For every Clone, a Drop can happen, and in this case, we want Drop to be called only once.

Fixes #238


Williangalvani (approved):

joaoantoniocardoso (commented):
Tested/validated with QGC:
image

PR #258, opened 2023-08-09
Support all raw formats

joaoantoniocardoso:


src/video/types.rs
```diff
@@ -5,6 +5,11 @@ use super::video_source_redirect::VideoSourceRedirect;
 use paperclip::actix::Apiv2Schema;
 use serde::{Deserialize, Serialize};
 
+pub const KNOWN_RTP_RAW_FORMATS: [&str; 9] = [
+    "RGB", "RGBA", "BGR", "BGRA", "AYUV", "UYVY", "I420", "Y41B", "UYVP",
+];
+pub const DEFAULT_RAW_FORMAT: &str = "I420";
```
patrickelectric, line 11: Did you look at https://docs.rs/gstreamer-video/latest/gstreamer_video/struct.VideoFormatInfo.html#method.name ?

Where the types are defined here:
https://docs.rs/gstreamer-video/latest/gstreamer_video/enum.VideoFormat.html

```
let info = VideoFormatInfo::from_format(crate::VideoFormat::RGB);
dbg!(info)'
```

src/video/types.rs
```diff
@@ -107,15 +109,64 @@ impl VideoSourceType {
...
-            "H264" => VideoEncodeType::H264,
-            "MJPG" => VideoEncodeType::Mjpg,
-            "YUYV" => VideoEncodeType::Yuyv,
-            _ => VideoEncodeType::Unknown(fourcc),
+
+        let fourcc = match fourcc.as_str() {
+            "H264" => return VideoEncodeType::H264,
+            "MJPG" => return VideoEncodeType::Mjpg,
+            // TODO: We can implement a [GstDeviceMonitor](https://gstreamer.freedesktop.org/documentation/gstreamer/gstdevicemonitor.html?gi-language=c) and then this manual mapping between v4l's and gst's fourcc will not be neccessary anymore.
+            // A list of possible v4l fourcc from the Linux docs:
```
patrickelectric, line 119: I believe that my previous comment shows how to do it with gstreamer VideoFormatInfo

src/video/types.rs
```diff
@@ -107,15 +109,64 @@ impl VideoSourceType {
...
+            "YUYV" => "YUY2".to_string(),
+            _ => fourcc,
+        };
+
+        // Use Gstreamer to recognize raw formats. [GStreamer raw formats](https://gstreamer.freedesktop.org/documentation/additional/design/mediatype-video-raw.html?gi-language=c#formats)
+        use std::ffi::CString;
+        if let Ok(c_char_array) = CString::new(fourcc.clone()).map(|c_str| c_str.into_raw()) {
+            use gst_video::ffi::*;
+
+            let gst_video_format = unsafe { gst_video_format_from_string(c_char_array) };
```
patrickelectric, line 132: I believe that my previous comment would also simplify this code.

src/video/types.rs
patrickelectric: And, also:
https://docs.rs/gstreamer-video/latest/gstreamer_video/enum.VideoFormat.html#method.to_fourcc

src/video/types.rs
```diff
@@ -107,15 +109,64 @@ impl VideoSourceType {
...
+                GST_VIDEO_FORMAT_UNKNOWN | GST_VIDEO_FORMAT_ENCODED
+            ) {
+                return VideoEncodeType::Raw(fourcc);
+            }
+        }
+
+        VideoEncodeType::Unknown(fourcc)
+    }
+}
+impl<'de> Deserialize<'de> for VideoEncodeType {
```
patrickelectric, line 145: please add a space between the implementations.

patrickelectric, line 145: Can you provide more information about why this implementation is necessary ? looking at the code this is not clear why.

patrickelectric (changes requested):
Some suggestions

src/video/types.rs
joaoantoniocardoso: > Because of the tricky nature of having two Enum fields both encapsulating Strings, plus to create a validation of what is accepted from GStreamer/V4l, I had to write a custom deserializer.

Like, if we define the `Unknown(String)` as `untagged`, everything different from the other options will be Unknown. If we then add another `untagged` `Raw(String)`, then it goes to who is defined first, and we actually want to use custom logic to define if it is Raw or Unknown.

patrickelectric: So, it would be nice to have such explanation as a comment.

Issue #260, opened 2023-08-10
Resource temporarily unavailable when configure stream


amerghazal7:
Hello there, 
After installing and building I tried to create a stream from my USB camera connected to the Jetson Orin, but nothing happens in the UI and I'm getting the error below in the console:

```
camera.rs:816: Error receiving a message as 1:100. Reason: Io(
    Os {
        code: 11,
        kind: WouldBlock,
        message: "Resource temporarily unavailable",
    },
). Camera: MavlinkCameraInformation {
    component: MavlinkCameraComponent {
        system_id: 1,
        component_id: 100,
        stream_id: 1,
        vendor_name: "Logitech BRIO",
        model_name: "/dev/video0 - Logitech BRIO",
        firmware_version: 0,
        resolution_h: 1080,
        resolution_v: 1920,
        framerate: 30.0,
        bitrate: 5000,
        rotation: 0,
        hfov: 90,
        thermal: false,
    },
    mavlink_connection_string: "tcpout:127.0.0.1:5760",
    mavlink_stream_type: VIDEO_STREAM_TYPE_RTPUDP,
    video_stream_uri: Url {
        scheme: "udp",
        cannot_be_a_base: false,
        username: "",
        password: None,
        host: Some(
            Domain(
                "0.0.0.0",
            ),
        ),
        port: Some(
            5600,
        ),
        path: "/logi",
        query: None,
        fragment: None,
    },
    video_source_type: Local(
        VideoSourceLocal {
            name: "Logitech BRIO",
            device_path: "/dev/video0",
            typ: Unknown(
                "usb-3610000.xhci-3.4",
            ),
        },
    ),
}

```

**NOTE** : there's no other apps or processes using the camera, and I'm using mavlink-router to provide a server for the mavlink communication.

Any idea how to solve that?

patrickelectric:
It appears that is failing to communicate with the mavlink server, are you running mavlink-router as tcpin ? Be sure to start the router before the camera manager.

amerghazal7:
I'm running the default config, this is the output for mavlink-server

![image](https://github.com/mavlink/mavlink-camera-manager/assets/105483346/7ea6b79b-93a1-4874-aeb6-603348b24fb0)

amerghazal7:
and the config is as below
```

[General]
   TcpServerPort=5760
   ReportStats=false
   MavlinkDialect=common

[UdpEndpoint local]
   Mode=normal
   Address=127.0.0.1
[...]

PR #262, opened 2023-08-10
src: stream: image_sink: Only set property if it exists

joaoantoniocardoso:
This prevents panicking when the `discard-corrupted-frames` property is not available (old GStreamer)


patrickelectric (approved):

patrickelectric:
May fix #260

PR #276, opened 2023-08-24
src: html: index: Avoid using absolute path for webrtc

patrickelectric:
When mcm is running under nginx or something else, the redirects will not work if we use absolute paths.


joaoantoniocardoso (approved):

joaoantoniocardoso:
hahahah that consistent rust formatter

Issue #273, opened 2023-08-22
Google Coral camera not detected on /dev/video3


steamfoundry:
The V4L compatible camera on /dev/video3 isn't detected currently.  

### Camera
 [Google Coral](https://coral.ai/products/camera/#tech-specs) connected via MIPI-CSI 

**Example Gstreamer command that will stream to QGroundControl (As UDP)**
sudo gst-launch-1.0 v4l2src device=/dev/video3 ! video/x-raw,width=640,height=480,framerate=30/1 ! vpuenc_h264 bitrate=500 ! rtph264pay ! udpsink host=192.168.1.103 port=5600 sync=false

### Board details

-     Board - NXP NavQPlus
-     Processor -NXP i.MX 8M Plus (4x Arm Cortex A53 cores)
-     OS: Ubuntu 22.04 on Yocto Linux 5.15 kernel - [Build 22.04.3 from Rudis Labs](https://github.com/rudislabs/navqplus-create3-images/releases/tag/v22.04.3)

Output from mavlink-camera-manager
```

2023-08-22T15:31:00.768533Z  INFO main ThreadId(01) src/logger/manager.rs:82: mavlink-camera-manager, version: 0.2.4-5e2a1f8, build date: 2023-08-21
2023-08-22T15:31:00.769195Z  INFO main ThreadId(01) src/logger/manager.rs:89: Starting at 2023-08-22T15:31:00
2023-08-22T15:31:00.769568Z DEBUG main ThreadId(01) src/logger/manager.rs:93: Command line call: ./mavlink-camera-manager -v --mavlink=udpout:10.0.0.2:14551 --enable-tracing-level-log-file
2023-08-22T15:31:00.769768Z DEBUG main ThreadId(01) src/logger/manager.rs:94: Command line input struct call: {
    "log-path": MatchedArg {
        occurs: 0,
        indices: [
            6,
        ],
        vals: [
            "./logs",
        ],
    },
    "enable-tracing-level-log-file": MatchedArg {
        occurs: 1,
        indices: [
            4,
        ],
        vals: [],
    },
    "rest-server": MatchedArg {
        occurs: 0,
        indices: [
            5,
        ],
        vals: [
            "0.0.0.0:6020",
        ],
    },
    "verbose": MatchedArg {
        occurs: 1,
        indices: [
            1,
        ],
        vals: [],
    },
    "mavlink": MatchedArg {
        occurs: 1,
        indices: [
            3,
        ],
        vals: [
[...]

Issue #139, opened 2022-10-07
Find a way to identify each camera instead of managing them by linux device path


joaoantoniocardoso:
The problem is that cameras may change from one USB port to another, and even when they are in the same USB port, there is no guarantee that their sub-devices will be created in the same order, resulting in `/dev/video0` with `H264` format on one boot, and maybe `/dev/video0` with `YUYV` on the next boot.

One idea to identify each camera and coherently keep the settings per camera (even when swapping between USB ports), is to not create streams based on the Linux device path when loading them from the settings file. Instead, we should use some UID like `ID_SERIAL` from `udevadm info --name=/dev/video0` combined with the device's format options, like H264, YUYV, etc.

PR #281, opened 2023-09-04
src: Get video formats using Device Monitor

joaoantoniocardoso:
Merge after #607 

### update 13/05/26:

This PR is the first step towards supporting libcamera. In this step, we move from the v4l crate to give us the stream formats to get them from the GStreamer DeviceMonitor.

The further steps are already planned, and libcamera will be functional in the next PR.

With these changes, the v4l crate is only used for the controls API (which is fine).

---

### update 09/29/25:

This seems very promising -- so far, I have already replaced the v4l formats and the v4l device discovery.

I'm thinking we should refactor the source/pipeline part of the application to fully make use of this Device Monitor:

1. I still need to reason about the architecture, but I think we can safely say that at least we should split our "Pipelines" (v4l2_pipeline.rs, etc) into a source (pipeline: source -> capsfilter -> parser -> proxysink) + transcoding pipeline (pipeline: proxysrc -> video_tee -> rtp_payloader -> capsfilter -> rtp_tee, as well as the future optional decoder+encoder), which will then connect to our many sink (pipeline: proxysrc -> sink).
2. Then we create a Device enum (w/ VideoSource and AudioSource variants) and a DeviceProvider trait, implemented by both this GSTDeviceMonitor and by the OnvifDiscovery, to give us a common interface.
3. Each "gst::Device" is also an element factory provider, which allows you to create its source element with no effort, which will help us to more easily reach libcamera and audio support.

---

Helps #372, #273, #139, #79

### DIFF
--- a/src/lib/stream/manager.rs
+++ b/src/lib/stream/manager.rs
@@ -17,7 +17,10 @@ use crate::{
         types::CaptureConfiguration,
         webrtc::signalling_protocol::BindAnswer,
     },
-    video::{types::VideoSourceType, video_source},
+    video::{
+        types::{Format, VideoSourceType},
+        video_source::{self, VideoSourceFormats},
+    },
     video_stream::types::VideoAndStreamInformation,
 };
 
@@ -151,8 +154,24 @@ pub async fn update_devices(
             continue;
         };
 
+        // Only resolve formats for candidates sharing the source name so that mismatched
+        // candidates (which are filtered out by name first) don't trigger device probing.
+        let mut formats: HashMap<String, Vec<Format>> = HashMap::new();
+        for candidate in candidates.iter() {
+            let VideoSourceType::Local(camera) = candidate else {
+                continue;
+            };
+            if camera.name != source.name {
+                continue;
+            }
+            formats.insert(
+                candidate.inner().source_string().to_string(),
+                candidate.formats().await,
+            );
+        }
+
         match source
-            .try_identify_device(capture_configuration, candidates)
+            .try_identify_device(capture_configuration, candidates, &formats)
             .await
         {
             Ok(Some(candidate_source_string)) => {
--- a/src/lib/video/gst_device_monitor.rs
+++ b/src/lib/video/gst_device_monitor.rs
@@ -1,0 +1,111 @@
+use std::sync::{Arc, Mutex};
+
+use anyhow::{Context, Result};
+use gst_app::prelude::*;
+use tracing::*;
+
+lazy_static! {
+    static ref MANAGER: Arc<Mutex<Manager>> = {
+        // Constructing `gst::DeviceMonitor` requires GStreamer to be initialized; ensure it is so
+        // that callers like cameras_available() work even when the binary entry point hasn't run.
+        gst::init().expect("Failed to initialize GStreamer");
+        Arc::new(Mutex::new(Manager::default()))
+    };
+}
+
+#[derive(Default)]
+struct Manager {
+    monitor: gst::DeviceMonitor,
+}
+
+impl Drop for Manager {
+    fn drop(&mut self) {
+        self.monitor.stop();
+    }
+}
+
+#[instrument(level = "debug")]
+pub fn init() -> Result<()> {
+    let manager_guard = MANAGER.lock().unwrap();
+
+    manager_guard.monitor.set_show_all_devices(true);
+    manager_guard.monitor.set_show_all(true);
+    manager_guard.monitor.start()?;
+
+    let providers = manager_guard.monitor.providers();
+    info!("GST Device Providers: {providers:#?}");
+
+    Ok(())
+}
+
+#[instrument(level = "debug")]
+pub(crate) fn video_devices() -> Result<Vec<glib::WeakRef<gst::Device>>> {
+    let monitor = &MANAGER.lock().unwrap().monitor;
+
+    let devices = monitor
+        .devices()
+        .iter()
+        .filter_map(|device| {
+            if device.device_class().ne("Video/Source") {
+                return None;
+            }
+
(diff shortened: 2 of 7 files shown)


src/video/video_source_local.rs
```diff
@@ -269,27 +259,200 @@ fn convert_v4l_intervals(v4l_intervals: &[v4l::FrameInterval]) -> Vec<FrameInter
...
+
+    let caps = gst_device_monitor::device_caps(&device)?;
+
+    let mut sizes_by_encode: HashMap<VideoEncodeType, HashSet<Size>> = HashMap::new();
+
+    caps.iter().for_each(|structure| {
+        let encode = match structure.name().as_str() {
+            "video/x-raw" => {
+                let fourcc = structure.get::<String>("format").unwrap();
+                VideoEncodeType::from_str(&fourcc)
```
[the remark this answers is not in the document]
joaoantoniocardoso: It doesn't help, "YUYV" is "YUY2" for GST regardless of creating the VideoFormat object.

patrickelectric:
@joaoantoniocardoso can you rebase this PR ?

joaoantoniocardoso:
@patrickelectric oops, accidentally clicked to merge before review; Can you review it and I make a following PR to adjust any discrepancies?

src/lib/video/gst_device_monitor.rs
```diff
@@ -0,0 +1,111 @@
...
+
+    manager_guard.monitor.set_show_all_devices(true);
+    manager_guard.monitor.set_show_all(true);
+    manager_guard.monitor.start()?;
+
+    let providers = manager_guard.monitor.providers();
+    info!("GST Device Providers: {providers:#?}");
+
+    Ok(())
+}
```
patrickelectric, line 39: just for curiosity, is there a reason for not having this in the default call of Manager ?

src/lib/video/gst_device_monitor.rs
```diff
@@ -0,0 +1,111 @@
...
+pub fn v4l_devices() -> Result<Vec<glib::WeakRef<gst::Device>>> {
+    let devices = video_devices()?
+        .iter()
+        .filter(|device_weak| {
+            let Some(device) = device_weak.upgrade() else {
+                return false;
+            };
+
+            device.properties().iter().any(|s| {
+                let Ok(api) = s.get::<String>("device.api") else {
```
patrickelectric, line 70: It would be nice to have a comment for this string:
```
# device.api is an inner string defined by gst for which api this device is accessed with: Ex: alsa, v4l..
```

It would be even better to have such strings as "device.path" to be defined as a global private property that explains that is a GST property string for this field

src/lib/video/local/video_source_local_linux.rs
```diff
@@ -719,70 +721,24 @@ impl VideoSource for VideoSourceLocal {
...
-        cameras
+        gst_device_monitor::v4l_devices()
+            .unwrap_or_default()
+            .iter()
+            .filter_map(|device_weak| {
+                let device = device_weak.upgrade()?;
+                let name = device.display_name().to_string();
+                let properties = device.properties()?;
+                let device_path = properties.get::<String>("device.path").ok()?;
+                let bus = properties.get::<String>("v4l2.device.bus_info").ok()?;
```
patrickelectric, line 732: isn't it better to have abstraction for it under gst_device_monitor as the function `device_caps` ?

src/lib/video/types.rs
```diff
@@ -29,34 +38,23 @@ pub enum VideoEncodeType {
     Yuyv,
 }
 
-#[derive(Apiv2Schema, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Deserialize, Serialize)]
```
patrickelectric, line 32: This commit is unnecessary moving this structure.

src/lib/stream/manager.rs
```diff
@@ -151,8 +154,24 @@ pub async fn update_devices(
...
+                continue;
+            };
+            if camera.name != source.name {
+                continue;
+            }
+            formats.insert(
+                candidate.inner().source_string().to_string(),
+                candidate.formats().await,
+            );
+        }
```
patrickelectric, line 171:
```suggestion
        for candidate in candidates.iter().filter(|candidate| {
            matches!(candidate, VideoSourceType::Local(camera) if camera.name == source.name)
        }) {
            formats.insert(
                candidate.inner().source_string().to_string(),
                candidate.formats().await,
            );
        }
```

src/lib/video/local/video_source_local_linux.rs
```diff
@@ -242,53 +245,301 @@ impl VideoSourceLocal {
...
+            // Yes, our nominator is GST's denominator.
+            numerator: value.denom() as u32,
+            denominator: value.numer() as u32,
+        }
+    }
+}
+
+fn get_device_formats_using_gstreamer(
+    device_path: &str,
+    _typ: &VideoSourceLocalType,
```
patrickelectric, line 289: why is this necessary ?

patrickelectric (commented):
I had no time to finish the review today. Will do the final commit tomorrow.

src/lib/video/types.rs
joaoantoniocardoso: I'm slowly moving things to get a top-to-bottom declaration, but if I added another commit, we'd have multiple commits touching the same lines.

src/lib/video/local/video_source_local_linux.rs
joaoantoniocardoso: The main idea is to use the DeviceMonitor's Device to interface the "local" kind of sources, and the previous approach to get formats was specific to v4l, so it wouldn't work for libcamerasrc or other sources.

PR #282, opened 2023-09-04
src: stream: webrtc: Move from async-std runtime to tokio

joaoantoniocardoso:


patrickelectric:
It's not clear from the commit body or the PR what is the intended reason behind this PR.

joaoantoniocardoso:
> It's not clear from the commit body or the PR what is the intended reason behind this PR.

The async-std runtime was being used in the example code and I forgot to translate this part after finishing the rest, so we are switching from creating async-std runtimes to creating them using the Tokio runtime. Also, from testing I realized that we don't need more than one worker, reducing the overhead.

patrickelectric (approved):

PR #283, opened 2023-09-12
Refactor mavlink

joaoantoniocardoso:
**TL;DR: No more WouldBlocks when accessing Mavlink TCP** :)

note: to be merged after #277 only


src/mavlink/manager.rs
```diff
@@ -1 +1,203 @@
...
+                Ok(message) => message,
+                Err(error) => {
+                    if let mavlink::error::MessageReadError::Io(io_error) = &error {
+                        if io_error.kind() == std::io::ErrorKind::WouldBlock {
+                            trace!("Blocked I/O, probably disconnected!");
+                            continue;
+                        }
+                    }
+                    error!("Failed receiving message from Mavlink Connection: {error:?}");
+                    break;
```
patrickelectric, line 91: Wait, should we truly break the loop ? Why not:
```suggestion
                    error!("Failed receiving message from Mavlink Connection: {error:?}");
                    std::thread::sleep(std::time::Duration::from_millis(500));
                    continue;
```

Is there a case that we want to truly break it ?

src/mavlink/manager.rs
```diff
@@ -1 +1,203 @@
...
+
+            // Send the received message to the cameras
+            if let Err(error) = inner
+                .lock()
+                .unwrap()
+                .sender
+                .send(Message::Received((header, message)))
+            {
+                error!("Failed handling message: {error:?}");
+                continue;
```
patrickelectric, line 105: this continue is unnecessary.

src/mavlink/manager.rs
```diff
@@ -1 +1,203 @@
...
+
+                // Receive answer from the cameras
+                let (header, message) = match receiver.try_recv() {
+                    Ok(Message::ToBeSent(message)) => message,
+                    Err(broadcast::error::TryRecvError::Closed) => {
+                        unreachable!(
+                            "Closed channel: This should never happen, this channel is static!"
+                        );
+                    }
+                    Ok(Message::Received(_)) | Err(_) => continue,
```
patrickelectric, line 126: Maybe:
```suggestion
                    something => {
                        warning!("Got something unexpected on the channel: {something:#?}");
                        continue;
                    }
```

src/mavlink/manager.rs
```diff
@@ -1 +1,203 @@
...
+                if let Err(error) = inner.lock().unwrap().connection.send(&header, &message) {
+                    error!("Failed sending message to Mavlink Connection: {error:?}");
+                    break;
+                }
+
+                trace!("Message sent: {header:?}, {message:?}");
+            }
+
+            // Reconnects
+            {
```
patrickelectric, line 139: The outside loop does not have a sleep, and that's super dangerous.
```suggestion
            {
                std::thread::sleep(std::time::Duration::from_millis(500));
```

src/mavlink/manager.rs
```diff
@@ -1 +1,203 @@
...
+
+    #[instrument(level = "debug")]
+    pub fn drop_id(id: u8) {
+        let manager = MANAGER.lock().unwrap();
+        let mut vector = manager.ids.lock().unwrap();
+
+        if let Some(position) = vector.iter().position(|&vec_id| vec_id == id) {
+            vector.remove(position);
+        } else {
+            error!("Id not found");
```
patrickelectric, line 170:
```suggestion
            error!("Id not found: {id}");
```

src/mavlink/manager.rs
```diff
@@ -1 +1,203 @@
...
+        connection.sender.clone()
+    }
+}
+
+impl Connection {
+    #[instrument(level = "debug")]
+    fn connect(address: &str) -> Box<dyn MavConnection<MavMessage> + Sync + Send> {
+        debug!("Connecting...");
+
+        loop {
```
patrickelectric, line 189:
```suggestion
        loop {
            debug!("Connecting to: {address}");
```

patrickelectric (changes requested):
1/3 of the review

src/mavlink/manager.rs
joaoantoniocardoso: Great, I'll put it inside the loop. 

Just so you know, when we add the tracing::instrument derive macro for a function, it instruments the function parameters in the log, so the address is already in the record :)
It will show like:
```
2023-09-14T16:24:05.013431Z DEBUG                 main ThreadId(01) connect{address="tcpout:0.0.0.0:5777"}: src/mavlink/manager.rs:188: Connecting...
```

src/mavlink/manager.rs
joaoantoniocardoso: Already being logged using tracing::instrument

src/mavlink/manager.rs
joaoantoniocardoso: Indeed there's not. If we break it, it's never going to recover anyway.

src/mavlink/manager.rs
joaoantoniocardoso: I agree, but if some day we decide to add extra lines below...

src/mavlink/manager.rs
joaoantoniocardoso: If the channel is working properly, the `Message::Received(_)` is expected to happen for every received message because we are sharing a single channel to both send and receive from/to between mavlink and our cameras. The other error is when the channel is empty, so we want to ignore both.

I'll add a comment to state that.

src/mavlink/manager.rs
joaoantoniocardoso: Placed right before the end of the context

src/mavlink/utils.rs
```diff
@@ -0,0 +1,129 @@
...
+
+pub fn from_string_to_vec_char_with_defined_size_and_null_terminator(
+    src: &str,
+    size: usize,
+) -> Vec<char> {
+    let mut uri = src.chars().collect::<Vec<char>>();
+    uri.truncate(size);
+    uri.push('\0');
+    uri
+}
```
joaoantoniocardoso, line 30: I was thinking about it and.. should it be:
```suggestion
pub fn from_string_to_vec_char_with_defined_size_and_null_terminator(
    src: &str,
    size: usize,
) -> Vec<char> {
    let mut uri = src.chars().collect::<Vec<char>>();
    uri.truncate(size - 1);
    uri.push('\0');
    uri
}
```

patrickelectric: The truncate does not make much sense to me. we should just add \0

joaoantoniocardoso: > The truncate does not make much sense to me. we should just add \0

The truncation is technical here: The protocol states, for example, 140 characters long, so our vector needs to have, at maximum, 140 characters long. We could create the vector with a capacity of 140 characters and copy each character into it, or to have this unsized Vec and then truncate it to the correct size.

patrickelectric: > > The truncate does not make much sense to me. we should just add \0
> 
> The truncation is technical here: The protocol states, for example, 140 characters long for a field, so our vector needs to have, at maximum, 140 characters long. We could create the vector with a capacity of 140 characters and copy each character into it, or, as it is being done here, to have this unsized Vec and then truncate it to the correct size (size of the original string, truncated at 140) after copying the characters into it.

Oh sorry, I thought that was str.size and not size as parameter. It makes totally sense. Truncate will limit the size of the array. To be safer, size -1 makes more sense indeed.

src/mavlink/mavlink_camera_component.rs
```diff
@@ -0,0 +1,79 @@
+use crate::video_stream::types::VideoAndStreamInformation;
+
+use anyhow::Result;
+use tracing::*;
+
+#[derive(Debug, Clone)] // Do not Clone this, otherwise it's very likely to have unwanted Drop calls
```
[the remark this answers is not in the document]
joaoantoniocardoso: Hah, good one. The comment was from the older implementation. It's safe now :D

src/mavlink/utils.rs
joaoantoniocardoso: Cool, instead of `truncate(lenght-1)`, I changed it to use `take(lenght-1)`.

patrickelectric (approved):

patrickelectric:
looks great!

PR #285, opened 2023-09-21
Fix thread leak

joaoantoniocardoso:
The thread leak was due to a reference to the webrtcbin element living inside the closures, keeping webrtcbin threads alive forever.

### DIFF
--- a/src/stream/manager.rs
+++ b/src/stream/manager.rs
@@ -40,7 +40,7 @@ pub struct Manager {
 }
 
 lazy_static! {
-    static ref MANAGER: Arc<Mutex<Manager>> = Arc::new(Mutex::new(Manager::default()));
+    static ref MANAGER: Arc<Mutex<Manager>> = Default::default();
 }
 
 impl Manager {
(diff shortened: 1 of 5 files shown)


src/stream/manager.rs
```diff
@@ -272,6 +271,14 @@ pub async fn get_jpeg_thumbnail_from_source(
...
         tokio::runtime::Builder::new_current_thread()
+            .on_thread_start(|| debug!("Thread started"))
+            .on_thread_stop(|| debug!("Thread stopped"))
+            .thread_name_fn(|| {
+                static ATOMIC_ID: std::sync::atomic::AtomicUsize =
+                    std::sync::atomic::AtomicUsize::new(0);
+                let id = ATOMIC_ID.fetch_add(1, std::sync::atomic::Ordering::SeqCst);
+                format!("Thumbnailer-{id}")
+            })
             .enable_time()
```
patrickelectric, line 282: Can we have a function to abstract such creation ? This code appears to duplicate alot.

joaoantoniocardoso: They are not exactly the same...

joaoantoniocardoso:
@patrickelectric 

Tested for 8 hours straight starting and ending 1 WebRTC session every 10 minutes, no thread leaks detected. In comparison, the current master would have had around 144 leaks with the same test.

patrickelectric (approved):

PR #286, opened 2023-09-22
Debug locks

joaoantoniocardoso:


joaoantoniocardoso:
See #326 for a more interesting approach to debugging the locks.

PR #287, opened 2023-09-22
Update libs

joaoantoniocardoso:
Already tested

### DIFF
--- a/src/video/xml.rs
+++ b/src/video/xml.rs
@@ -13,6 +13,7 @@ pub struct MavlinkCamera {
 
 #[derive(Debug, Serialize)]
 pub struct Definition {
+    #[serde(rename = "@version")]
     pub version: u32,
     pub model: Model,
     pub vendor: Vendor,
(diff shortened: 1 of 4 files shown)


src/video/xml.rs
```diff
@@ -56,33 +57,45 @@ pub enum ParameterType {
 
 #[derive(Debug, Serialize)]
 pub struct ParameterBool {
+    #[serde(rename = "@name")]
     pub name: String,
-    #[serde(rename = "type")]
+    #[serde(rename = "@type")]
     pub cpp_type: String,
+    #[serde(rename = "@default")]
```
patrickelectric, line 64: why all this @ are necessary ?

joaoantoniocardoso: They change the serializer behavior to use the fields as parameters (as `<tag paramter="value"/> instead of sub-tags (as `<tag><subtag/></tag>`).

joaoantoniocardoso: Done

patrickelectric (approved):

PR #293, opened 2023-10-24
Better debug

joaoantoniocardoso:


src/stream/sink/image_sink.rs
```diff
@@ -57,23 +57,33 @@ impl CachedThumbnails {
...
     _proxysrc: gst::Element,
+    #[derivative(Debug = "ignore")]
     _transcoding_elements: Vec<gst::Element>,
+    #[derivative(Debug = "ignore")]
     appsink: gst_app::AppSink,
+    #[derivative(Debug = "ignore")]
     tee_src_pad: Option<gst::Pad>,
+    #[derivative(Debug = "ignore")]
     flat_samples_sender: tokio::sync::broadcast::Sender<ClonableResult<FlatSamples<Vec<u8>>>>,
+    #[derivative(Debug = "ignore")]
```
joaoantoniocardoso, line 80: This is ugly, maybe we could change it to an internal struct, like:

```Rust
struct Inner {
    pipeline: gst::Pipeline,
    queue: gst::Element,
    proxysink: gst::Element,
    _proxysrc: gst::Element,
    _transcoding_elements: Vec<gst::Element>,
    appsink: gst_app::AppSink,
    tee_src_pad: Option<gst::Pad>,
}

#[derive(derivative::Derivative)]
#[derivative(Debug)]
pub struct ImageSink {
    sink_id: uuid::Uuid,
    #[derivative(Debug = "ignore")]
    inner: Inner,
    flat_samples_sender: tokio::sync::broadcast::Sender<ClonableResult<FlatSamples<Vec<u8>>>>,
    pad_blocker: Arc<Mutex<Option<gst::PadProbeId>>>,
    pipeline_runner: PipelineRunner,
    thumbnails: Arc<Mutex<CachedThumbnails>>,
}
```

patrickelectric: What are the consequences of such approach ? If there are none, we could just change the code.

PR #313, opened 2023-11-29
mavlink: mavlink_camera: Add debug message then removing mavlink camera handle

patrickelectric:


joaoantoniocardoso:
Superposed by #330

Issue #310, opened 2023-11-28
Error creating thread: Resource temporarily unavailable


joaoantoniocardoso:
Related report:
> Another crash
> Also something has changed with stream thumbnails.. they pop in and out a lot more now. I get the message
> Thumbnail for source ""ball"" is temporarily unavailable. Try again later. Details: deadline has elapsed""""
> Also happened on a customers ROV right now. They were at 60m depth and i had to go in and restart the MCM service.
Here are the recent logs as well as terminal output.

Possible thread leak, leading to SO thread limit, but which threads? From a brief investigation, it seems to be Mavlink and Signaller.

```
2023-11-21T09:29:36.117533Z DEBUG                                     MavlinkCamera-9 ThreadId(1375) heartbeat_loop{component_id=100}: src/mavlink/mavlink_camera.rs:159: Heartbeat sent

(mavlink-camera-manager:79): GStreamer-WARNING **: 09:29:36.347: failed to create thread: Error creating thread: Resource temporarily unavailable
2023-11-21T09:29:36.996633Z DEBUG                                         Signaller-1 ThreadId(06) accept_connection{peer=192.168.40.233:42002}:handle_connection{peer=192.168.40.233:42002}:handle_message{msg=Text(""{\""type\"":\""question\"",\""content\"":{\""type\"":\""startSession\"",\""content\"":{\""consumer_id\"":\""c891ea9f-18d1-42a9-9e4e-22c9c6a7c696\"",\""producer_id\"":\""b8bdeb99-07b0-42a6-989f-9e84105096e6\""}}}"") sender=UnboundedSender { chan: Tx { inner: Chan { tx: Tx { block_tail: 0x24f3d8f0, tail_position: 2 }, semaphore: Semaphore(0), rx_waker: AtomicWaker, tx_count: 1, rx_fields: ""..."" } } }}:add_session{bind=BindOffer { consumer_id: c891ea9f-18d1-42a9-9e4e-22c9c6a7c696, producer_id: b8bdeb99-07b0-42a6-989f-9e84105096e6 } sender=UnboundedSender { chan: Tx { inner: Chan { tx: Tx { block_tail: 0x24f3d8f0, tail_position: 2 }, semaphore: Semaphore(0), rx_waker: AtomicWaker, tx_count: 2, rx_fields: ""..."" } } }}: src/stream/manager.rs:430: WebRTC session created: dc94a03d-7175-4388-9985-7ac947511dbc
[...]

PR #314, opened 2023-11-29
mavlink: mavlink_camera: Add shutdown mechanism

patrickelectric:
Without it, the tasks never finishes properly

May help #310


joaoantoniocardoso:
Cool, this trigger can indeed solve the problem. In this case, it would be better to have a ReadWriteLock, or an atomic bool

joaoantoniocardoso:
Superposed by #330

PR #315, opened 2023-11-29
mavlink: Link the life cycle of the threads to the channel

joaoantoniocardoso:
Replaces #314

May help #310

### DIFF
--- a/src/mavlink/mavlink_camera.rs
+++ b/src/mavlink/mavlink_camera.rs
@@ -152,8 +152,8 @@ impl MavlinkCamera {
             });
 
             if let Err(error) = sender.send(Message::ToBeSent((header, message))) {
-                error!("Failed to send message: {error:?}");
-                continue;
+                debug!("Finishing thread. Reason: {error:?}");
+                break;
             }
 
             debug!("Heartbeat sent");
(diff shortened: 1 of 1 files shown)


src/mavlink/mavlink_camera.rs
```diff
@@ -607,15 +600,21 @@ impl MavlinkCamera {
...
+        res
     }
 }
 
 impl Drop for MavlinkCameraHandle {
     fn drop(&mut self) {
-        self.heartbeat_handle.abort();
-        self.messages_handle.abort();
-        super::manager::Manager::drop_id(self.inner.component.component_id)
+        if !self.heartbeat_handle.is_finished() || !self.messages_handle.is_finished() {
```
patrickelectric, line 613: It does not fix the issue compared with #314.
Change this if to a while and you'll see.

image

patrickelectric (changes requested):

src/mavlink/mavlink_camera.rs
joaoantoniocardoso: Thanks. here that doesn't happen... Then I'm afraid I can't effectively validate the other as well.

PR #317, opened 2023-11-30
Add thread counter function

patrickelectric:


patrickelectric:
@joaoantoniocardoso sadly the counter is returning zero, not sure why since the threads are processes that run with the current thread as their parent id. Any idea ?

patrickelectric:
> @joaoantoniocardoso sadly the counter is returning zero, not sure why since the threads are processes that run with the current thread as their parent id. Any idea ?

@joaoantoniocardoso I got this working now and I'm able to identify thread leaks. 
1. Enable the monitor
2. Create a stream
3. Go to webrtc site and do all the steps to see the video stream
4. Repeat step 3 and you'll see the number of tasks increasing non stop

joaoantoniocardoso (approved):

PR #303, opened 2023-11-22
src: stream: Post EOS from a new thread

joaoantoniocardoso:
Fixes #309

PR #318, opened 2023-11-30
Better tasking less locks

joaoantoniocardoso:
merge after #303

### DIFF
--- a/src/server/manager.rs
+++ b/src/server/manager.rs
@@ -27,9 +27,8 @@ pub async fn run(server_address: &str) -> Result<(), std::io::Error> {
         App::new()
             // Add debug call for API access
             .wrap_fn(|req, srv| {
-                trace!("{:#?}", &req);
-                let fut = srv.call(req);
-                async { fut.await }
+                trace!("{req:#?}");
+                srv.call(req)
             })
             .wrap(TracingLogger::default())
             .wrap(actix_web::middleware::Logger::default())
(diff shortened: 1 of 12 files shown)


src/server/manager.rs
```diff
@@ -67,25 +66,26 @@ pub async fn run(server_address: &str) -> Result<(), std::io::Error> {
...
+            //                 InMemoryBackend::builder().build(),
+            //                 SimpleInputFunctionBuilder::new(std::time::Duration::from_secs(1), 4)
+            //                     .real_ip_key()
+            //                     .build(),
+            //             )
+            //             .add_headers()
+            //             .build(),
+            //         )
+            //         .route("", web::get().to(pages::thumbnail)),
+            // )
```
joaoantoniocardoso, line 83: This needs to be worked with care to work with only one worker...

src/server/pages.rs
```diff
@@ -380,41 +382,41 @@ pub fn sdp(sdp_file_request: web::Query<SdpFileRequest>) -> HttpResponse {
...
+//         Some(Err(error)) => HttpResponse::ServiceUnavailable()
+//             .reason("Thumbnail temporarily unavailable")
+//             .insert_header((header::RETRY_AFTER, 10))
+//             .content_type("text/plain")
+//             .body(format!(
+//             "Thumbnail for source {:?} is temporarily unavailable. Try again later. Details: {error:?}",
+//             thumbnail_file_request.source
+//         )),
+//     }
+// }
```
joaoantoniocardoso, line 422: This needs to be reworked with care to work with only one worker...

src/stream/mod.rs
```diff
@@ -154,27 +156,33 @@ impl Stream {
...
+                let new_state = match StreamState::try_new(
+                    &video_and_stream_information,
+                    &pipeline_id,
+                )
+                .await
+                {
+                    Ok(state) => state,
+                    Err(error) => {
+                        error!("Failed to recreate the stream {pipeline_id:?}: {error:#?}. Trying again in one second...");
+                        std::thread::sleep(std::time::Duration::from_secs(1));
```
joaoantoniocardoso, line 174:
```suggestion
                        tokio::time::sleep(tokio::time::Duration::from_secs(1));
```

patrickelectric:
> > The threads do not finish..
> > image
> > A shutdown mechanism like #314 is still necessary
> 
> Even if you add the shutdown mechanism on top of this patch, your test will fail. If you have a while loop inside the drop, you need to allow more workers.

Indeed the method does not work with this approach, I'll check if there other ways for us to investigate this.

joaoantoniocardoso:
Superposed by #377

PR #325, opened 2023-12-01
readme: Add missing dependency

joaoantoniocardoso:
libgstreamer1.0-x is required for pango, required for timeoverlay used on our fake streams


patrickelectric (approved):
Necessary for timeoverlay element

PR #326, opened 2023-12-01
[WIP] Investigation on leaks

joaoantoniocardoso:
With this group of commits, it seems that the leaks are from inside GST. Maybe we are not finishing it correctly, maybe we are keeping references around, like a pad/element/pipeline clone


joaoantoniocardoso:
Closed as all interesting parts of the work were already merged.

Issue #308, opened 2023-11-28
To investigate - sequence of problems when recreating streams


joaoantoniocardoso:
1. Both streams seem to have ended at the same time. Is this a USB problem on Linux?
2. Maybe because of timing, but the devices are switched (`/dev/video2` is recreated as `/dev/video6`.
3. Somehow, at some point we try to create a null device ( `device_path: """"`)
4. What is StateChangeError?

```
[...]

PR #330, opened 2024-01-06
Fix stream recreation crashes and thread starvation caused by MAVLink-related thread leak

joaoantoniocardoso:
note: **Merge after #303**

This work:
- Fix critical crashes when recreating streams
- Fix MAVlink-related thread leak that leads to a crash or complete stall caused by thread starvation
- Helps #310 because it fixes a thread leak related to the Mavlink and stream recreation
- May helps #308


patrickelectric (approved):
Tested and working

Issue #274, opened 2023-08-23
Support transcoding


amerghazal7:
Hello, 
I was wondering If we can add the ability to stream different media types not just the native one that the camera provides, as you are relying on the gstreamer isn\t that possible? Like for example I can take any input format that my camera support, and output stream in any codec like h264.
If that's possible but not implemented, I would really appreciate a guide so I can add it on my own.

Thanks

joaoantoniocardoso:
Welcome back @amerghazal7! First, thanks for the suggestion, and secondly, it is definitely possible, we just need to decide on a good path to do so.

### Pipeline

Before anything, let's see how it is currently working:

When a user creates a UDP Sink from a v4l2 USB camera, a "v4l pipeline" is created with a Tee in its end, then, two sink pipelines are sequentially connected to the Tee using queues: the "UDP Sink" and the "Image Sink". After that, whenever a WebRTC client requests a stream, a new "WebRTC Sink" is also connected to the Tee, ending with a pipeline as in the following image for 1 WebRTC client:

![image](https://github.com/mavlink/mavlink-camera-manager/assets/5920286/d7ac38e7-071a-400e-b67f-c45824a3b10b)

Now, If we don't want to leave the original format (from the v4l source) available for anything, a simple version could be implemented as so: we add three optional elements (in red) to do the transcoding if the transcoding is enabled, exposing three new fields to the user to configure: decoder, encoder, and the final format, as shown in the image below:

![image](https://github.com/mavlink/mavlink-camera-manager/assets/5920286/ab02b139-b030-4301-bd8d-375b68f8c58e)

If more performance is needed, we could get rid of some heavy elements by adding other Tee elements in strategic places, however, the complexity of managing it can grow considerably:

![image](https://github.com/mavlink/mavlink-camera-manager/assets/5920286/eee61798-2492-4d47-9707-81035ddf53a4)

### API

I prefer to let the user choose the specific decoder and encoder elements he wants to use because I don't want to let the GStreamer arbitrarily choose (using the videodecoder / videoencoder elements). For this, we can use GStreamer to list all possible decoders and decoders to do the transcoding from the original format (the v4l format) to the desired format (the sink format). Also, we can expose all the encoder/decoder properties so we don't need to manually add support for each possible encoder/decoder. Optionally, these properties could be exposed as Video/Stream Controls so could be changed in runtime.

---

@amerghazal7 and @patrickelectric Let me know what you think about it, especially the API.

Thanks

[...]

PR #331, opened 2024-01-12
Add video tee

joaoantoniocardoso:
This patch removes the need for some sinks (ImageSink for now, WebRTCSink in the future) to have to un an RTP depayloader and/or video parser.

It does so by adding a second Tee just before the payloader.

The older Tee, previously called "SinkTee", is now named "RTPTee", while the newer is called "VideoTee".

image

Relates to #274.


patrickelectric (approved):
nice!

PR #332, opened 2024-01-12
Improve queues and shm

joaoantoniocardoso:


patrickelectric (commented):
joao, can you give more details on that ? is-live and do-timestamp related to sync and wait-for-connection ? The commit title is not clear with the code.

patrickelectric:
@joaoantoniocardoso can you reply my previous comment ?

joaoantoniocardoso:
> @joaoantoniocardoso can you reply my previous comment ?

> joao, can you give more details on that ? is-live and do-timestamp related to sync and wait-for-connection ? The commit title is not clear with the code.

The sync and _do-timestamp_ are a combination to use the timestamps from the source, and _wait-for-connection_ seems to be required for that.

patrickelectric (approved):

PR #341, opened 2024-02-07
Allow MAVLink System ID to be specified via CLI argument

### LINKED ISSUE #327: Allow configuring system id
It should be set via an endpoint, as well command line argument, and stored in the config file. When changed, it should reflect to each mavlink component, better if we don't need to reconstruct them, so maybe an `Arc<RwLock<u16>>`.

https://github.com/mavlink/mavlink-camera-manager/blob/42109b37899c982021ecf4060096d48b02ddf586/src/mavlink/mavlink_camera_component.rs#L51

joaoantoniocardoso:
Closes #327

### DIFF
--- a/src/cli/manager.rs
+++ b/src/cli/manager.rs
@@ -73,6 +73,10 @@ struct Args {
     /// Enable webrtc thread test with limit of child tasks (can use port for webdriver as parameter).
     #[arg(long, value_name = "PORT", num_args = 0..=1, default_missing_value = "9515")]
     enable_webrtc_task_test: Option<u16>,
+
+    /// Sets the MAVLink System ID.
+    #[arg(long, value_name = "SYSTEM_ID", default_value = "1")]
+    mavlink_system_id: u8,
 }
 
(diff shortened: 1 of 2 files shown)


src/cli/manager.rs
```diff
@@ -73,6 +73,10 @@ struct Args {
     /// Enable webrtc thread test with limit of child tasks (can use port for webdriver as parameter).
     #[arg(long, value_name = "PORT", num_args = 0..=1, default_missing_value = "9515")]
     enable_webrtc_task_test: Option<u16>,
+
+    /// Sets the MAVLink System ID.
+    #[arg(long, value_name = "SYSTEM_ID", default_value = "1")]
+    mavlink_system_id: u8,
```
joaoantoniocardoso, line 79: @patrickelectric is all u8 range valid, or should we do some validation?

patrickelectric: I see no reason to not use the full range.

patrickelectric (approved):

PR #344, opened 2024-02-09
src: stream: sink: Add FailSafeKiller thread

joaoantoniocardoso:


patrickelectric (changes requested):
Not working, it also appears to freeze the entire process

joaoantoniocardoso:
Surpassed by #385

PR #362, opened 2024-03-14
Fix stream recreation logic, prefer Interval::tick to time::sleep in tokio loops

joaoantoniocardoso:


src/stream/pipeline/runner.rs
```diff
@@ -166,41 +166,44 @@ impl PipelineRunner {
         let mut lost_timestamps: usize = 0;
         let max_lost_timestamps: usize = 30;
 
-        loop {
-            tokio::time::sleep(tokio::time::Duration::from_millis(100)).await;
-
-            if let Some(reason) = finish.recv().await {
-                return Err(anyhow!("{reason:?}"));
-            }
```
joaoantoniocardoso, line 174: If not clear, the wrong logic was here. We were only going to the if below (which triggers the stream recreation process) after this future.

patrickelectric (approved):

PR #370, opened 2024-03-26
Add default argument for mavlink

patrickelectric:


joaoantoniocardoso (approved):
nice!

joaoantoniocardoso:
Oh, there is a settings test breaking because of the non-optional?

PR #424, opened 2024-07-15
Add windows and mac deployment with bun

patrickelectric:


patrickelectric:
@joaoantoniocardoso it appears that the mac build is being limited by the organization limit.

.github/workflows/deploy-unix.yml
```diff
@@ -21,11 +21,21 @@ jobs:
             TARGET: aarch64-unknown-linux-gnu
             SUFFIX: aarch64
 
+          - os: macos-13-xlarge
+            TARGET: aarch64-apple-darwin
+            SUFFIX: macos-aarch64
+
+          - os: windows-latest
```
joaoantoniocardoso, line 28:
```suggestion
          - os: windows-2022
```

.github/workflows/deploy-unix.yml
```diff
@@ -21,11 +21,21 @@ jobs:
             TARGET: aarch64-unknown-linux-gnu
             SUFFIX: aarch64
 
+          - os: macos-13-xlarge
```
joaoantoniocardoso, line 24:
```suggestion
          - os: macos-13
```

joaoantoniocardoso: large runners require enterprise stuff

.github/workflows/deploy-unix.yml
joaoantoniocardoso: `latest` is always out of out control, then it updates and the build suddenly starts breaking

build.rs
```diff
@@ -82,49 +99,49 @@ fn generate_typescript_bindings() {
...
         .status()
         .unwrap();
 
     #[cfg(debug_assertions)]
-    let install = Command::new("yarn")
+    let install = Command::new("bun")
         .args(["install", "--frozen-lockfile"])
         .current_dir(frontend_dir)
         .status()
         .unwrap();
```
joaoantoniocardoso, line 133: why do we have these two of the same? haha

patrickelectric:
@joaoantoniocardoso updated

joaoantoniocardoso (approved):

PR #441, opened 2024-10-08
src: lib: stream: sink: Set RTP Storage to 1 second

joaoantoniocardoso:
The default is 50ms


patrickelectric (commented):
It would be nice to have in the commit body the explaination for this change

Williangalvani:
Testing it and while I can't quantify any performance gain, it doesn't seem to have any performance impact, my cpu is sitting at 23% while running ardusub and streaming 4k video.

PR #459, opened 2024-12-11
Improve Redirect to WebRTC compatibility

joaoantoniocardoso:
Since WebRTC only works with a narrow option of stream formats, we are now re-encoding the RTP packages for Redirect streams, using the same strategy as for the Onvif sources.

An example is the stream from the Swell Pro. It's a 720p H264 stream but with an RTP incompatible with WebRTC.

### DIFF
--- a/src/lib/stream/gst/utils.rs
+++ b/src/lib/stream/gst/utils.rs
@@ -1,5 +1,6 @@
 use anyhow::{anyhow, Result};
 use gst::prelude::*;
+use tokio::sync::mpsc;
 use tracing::*;
 
 use crate::video::types::VideoEncodeType;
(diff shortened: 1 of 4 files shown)


src/lib/stream/gst/utils.rs
```diff
@@ -110,32 +111,63 @@ pub async fn wait_for_element_state_async(
...
+    };
 
     let pipeline = gst::parse::launch(&description)
         .expect("Failed to create pipeline")
         .downcast::<gst::Pipeline>()
         .expect("Pipeline is not a valid gst::Pipeline");
 
+    let typefinder = pipeline.by_name("typefinder")?;
+
+    let (tx, rx) = mpsc::channel(10);
```
patrickelectric, line 152: A oneshot makes more sense here, no ?

joaoantoniocardoso: That was my first try, but oneshot requires T to impl copy  :(

patrickelectric (approved):

PR #465, opened 2025-03-04
feature: keep config persistent after reset or restart if the stream is unavailable during restart

YehorFedorov:
Once I added the restart_streams API I found that manager::start_default function has the next line
```
    // Gently remove all streams as we are going to replace the entire list below
    remove_all_streams().await?;
```
Which was removing the full configuration. After a restart, the config contains the streams that were available during the restart and other disappeared

`remove_all_streams` is only called from `start_default`, so I decided to resolve this simply by providing a flag, that allow or prevent the config file from modification

What do you think?

### DIFF
--- a/src/lib/stream/manager.rs
+++ b/src/lib/stream/manager.rs
@@ -96,7 +96,7 @@ pub async fn remove_all_streams() -> Result<()> {
     };
 
     for stream_id in keys.iter() {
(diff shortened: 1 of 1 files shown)


joaoantoniocardoso (approved):
Makes sense, thanks!

PR #468, opened 2025-03-07
Rework streams creation

### LINKED ISSUE #345: Rapidly fired stream creation requests can result in duplicated streams
This is an architectural. There should only be one stream request being attended at each time, so we need to place the mutex in the right place.

### LINKED ISSUE #466: Redirect streams are being forgotten
How to reproduce it:

1. Configure a redirected stream
2. Power-cycle the system (with the camera)
3. See no configuration

It's probably because the RTSP stream is not available when MCM is creating streams, then it invalidates the stream and removes it. We should just flag it somehow instead of removing it from the configuration.

joaoantoniocardoso:
- Closes #466 
- Closes #345

By adding a stream, the system first adds it to the manager, and then the watcher task keeps trying to actually create the stream.

The downside is that the user won't have an immediate error returning from the stream creation, and he would have to check the `running` state. To address this, an optional "error" field was added to the StreamStatus so the user can have feedback on what is going on inside the stream creation.

## In case of error:

image

## In case of success:

image

### DIFF
--- a/src/lib/stream/mod.rs
+++ b/src/lib/stream/mod.rs
@@ -10,9 +10,9 @@ use std::sync::Arc;
 
 use ::gst::prelude::*;
 use anyhow::{anyhow, Context, Result};
-use gst::utils::get_encode_from_stream_uri;
+use gst::utils::get_capture_configuration_from_stream_uri;
 use manager::Manager;
-use pipeline::Pipeline;
+use pipeline::{Pipeline, PipelineGstreamerInterface};
 use sink::{create_image_sink, create_rtsp_sink, create_udp_sink};
 use tokio::sync::RwLock;
 use tracing::*;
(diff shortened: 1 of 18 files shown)


src/lib/stream/mod.rs
```diff
@@ -37,44 +37,37 @@ use self::{
...
-
-                let Some(encode) = get_encode_from_stream_uri(&url).await else {
-                    return Err(anyhow!("No encode found for stream"));
-                };
-
                 video_and_stream_information
                     .stream_information
                     .configuration = CaptureConfiguration::Video(VideoCaptureConfiguration {
-                    encode,
+                    encode: VideoEncodeType::Unknown("?".to_string()),
```
patrickelectric, line 70: maybe "Redirect stream" or something like it would make it more clear than just an interrogation

patrickelectric (approved):

patrickelectric:
@joaoantoniocardoso can you rebase over master ?

joaoantoniocardoso:
@patrickelectric am I missing something?

image

patrickelectric:
> @patrickelectric am I missing something?
> 
> image

Now it's ok.

PR #470, opened 2025-03-08
src: lib: stream: sink: Simplify Sinks link/unlink

joaoantoniocardoso:


src/lib/stream/sink/mod.rs
```diff
@@ -99,3 +100,166 @@ pub fn create_image_sink(
...
+    elements: &[&gst::Element],
+) -> Result<()> {
+    elements
+        .windows(2)
+        .enumerate()
+        .try_for_each(|(position, pair)| {
+            let src_element = pair[0];
+            let sink_element = pair[1];
+
+            // let src_pad = &src_element.static_pad("src").unwrap();
```
patrickelectric, line 178: are these comments necessary ?

patrickelectric (changes requested):

src/lib/stream/sink/mod.rs
joaoantoniocardoso: definitely not, ty!

joaoantoniocardoso:
@patrickelectric commented code was removed, rebased over master, and removed an unused property in RTSPSink.

patrickelectric (approved):
nice cleanup

PR #471, opened 2025-03-10
src: lib: video: local: Reduce V4L verbosity

joaoantoniocardoso:
Every time some client requests the available streams, MCM throws all these logs:

image

Since RadCam Manager is a client that requests that every second, the size of the logs is getting huge.

This PR addresses that by making those logs only available via trace.

### DIFF
--- a/src/lib/video/local/video_source_local_linux.rs
+++ b/src/lib/video/local/video_source_local_linux.rs
@@ -67,6 +67,7 @@ impl VideoSourceLocalType {
     //
     // https://www.kernel.org/doc/html/v4.9/media/uapi/v4l/vidioc-querycap.html#:~:text=__u8-,bus_info,-%5B32%5D
 
+    #[instrument(level = "debug")]
     pub fn from_str(description: &str) -> Self {
         if let Some(result) = VideoSourceLocalType::usb_from_str(description) {
             return result;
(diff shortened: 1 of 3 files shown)


src/lib/video/local/video_source_local_linux.rs
```diff
@@ -238,6 +247,8 @@ impl VideoSourceLocal {
     }
 }
 
+#[instrument(level = "debug")]
+
```
patrickelectric, line 251: empty line after macro

src/lib/video/local/video_source_local_linux.rs
```diff
@@ -284,6 +295,8 @@ fn convert_v4l_intervals(v4l_intervals: &[v4l::FrameInterval]) -> Vec<FrameInter
     intervals
 }
 
+#[instrument(level = "debug")]
+
```
patrickelectric, line 299: empty line after macro

src/lib/video/local/video_source_local_linux.rs
```diff
@@ -67,6 +67,8 @@ impl VideoSourceLocalType {
     //
     // https://www.kernel.org/doc/html/v4.9/media/uapi/v4l/vidioc-querycap.html#:~:text=__u8-,bus_info,-%5B32%5D
 
+    #[instrument(level = "debug")]
+
```
patrickelectric, line 71: empty line after macro

src/lib/video/local/video_source_local_linux.rs
```diff
@@ -201,6 +207,8 @@ impl VideoSourceLocal {
             .collect()
     }
 
+    #[instrument(level = "debug")]
+
```
patrickelectric, line 211: empty line after macro

patrickelectric (changes requested):

PR #475, opened 2025-03-13
src: lib: stream: sink: Add missing set_state to UDP and Image sinks

joaoantoniocardoso:
Oops, I missed this part: we need to set the state to Null of any pipeline created by the sink, otherwise, that pipeline lives forever.


joaoantoniocardoso (commented):
All tested, gonna merge it and release a patched version.

PR #485, opened 2025-05-13
created updated Dockerfile

ttb-git:


patrickelectric (commented):
Ubuntu 24.04 is a LTS, if changes are necessary here, I would vote to move the docker to use alpine to decrease the size.

PR #499, opened 2025-06-08
Render dot streams in the frontend

patrickelectric:
I would prefer to use egui on the front, but oh well... html it's 

image

### DIFF
--- a/src/lib/server/pages.rs
+++ b/src/lib/server/pages.rs
@@ -6,8 +6,12 @@ use actix_web::{
     web::{self, Json},
     HttpRequest, HttpResponse,
 };
+use actix_ws::Message;
+use futures::StreamExt;
 use paperclip::actix::{api_v2_operation, Apiv2Schema, CreatedJson};
 use serde::{Deserialize, Serialize};
+use serde_json::json;
+use tokio::time::Duration;
 use tracing::*;
 use validator::Validate;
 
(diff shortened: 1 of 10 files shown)


patrickelectric:
> Awesome feature!!!
> 
> 1. This feature will only work with internet access, but it's okay for now
> 2. This feature is getting the DOT for the "source" part of the pipeline, but we also want every single "sink" part connected to it.

Any idea of how to get the number 2 ?

src/lib/server/pages.rs
```diff
@@ -541,3 +545,68 @@ pub async fn unauthenticate_onvif_device(
...
+                            for stream_info in streams {
+                                if let Some((dot, childrens)) = crate::stream::manager::Manager::get_stream_dot_by_id(&stream_info.id).await {
+                                    let id = stream_info.id.to_string();
+                                    if last_dots.get(&id).map_or(true, |last_dot| last_dot != &dot) {
+                                        last_dots.insert(id.clone(), dot.clone());
+                                        changed = true;
+                                        dots.push(json!({
+                                            "id": id,
+                                            "dot": dot,
+                                            "childrens": childrens,
```
joaoantoniocardoso, line 582: "children" is already plural :sweat_smile: 
```suggestion
                                            "children": children,
```

src/lib/server/pages.rs
```diff
@@ -541,3 +545,68 @@ pub async fn unauthenticate_onvif_device(
...
+                                        }));
+                                    }
+                                }
+                            }
+                        }
+                        Err(e) => {
+                            warn!("Failed to get streams information: {}", e);
```
joaoantoniocardoso, line 589:
```suggestion
                        Err(error) => {
                            warn!("Failed to get streams information: {error:?}");
```

joaoantoniocardoso (approved):
Feel free to merge after considering the suggestions

PR #530, opened 2025-09-18
Improve pipeline runner

joaoantoniocardoso:
This patch:
- Fixes unnecessary stream recreation of low-fps sources
- Reduces wrongly identifying freezes for some streams (depending on the encoding, the position _does not_ change at every frame)


joaoantoniocardoso:
edit: simplified the approach after extensive testing. There are better ways of doing what we are doing here, such as pad probing, but these changes are already a significant gain.

joaoantoniocardoso:
@patrickelectric This is ready for review!

patrickelectric (approved):

PR #546, opened 2025-09-30
src: lib: stream: gst: utils: Filter out non-video channels in rtspsrc

joaoantoniocardoso:
This improves our success rate, as sometimes it was choosing the audio channel :disappointed:


patrickelectric:
@joaoantoniocardoso can you rebase ?

PR #569, opened 2026-02-14
do not retransmit

patrickelectric:
should be merged after #568


rafaellehmkuhl:
> Since we are running via the wire we are not losing packages in the rtsp connection, different from wifi. This approach reduce the cpu usage, including using tcp over udp to the camera connection.

Are we sure about that? We see high numbers on packet loss and retransmission all the time.

joaoantoniocardoso:
The retransmission is now disabled (since #593). In the future, we might want to allow that option.

PR #593, opened 2026-03-26
[next] [5] Rework stream architecture with lifecycle state machine and latency optimizations

joaoantoniocardoso:
> Note: The dedicated WebRTC leak workflow is removed here; broader browser/integration coverage is added back on #594.

## Summary

Major rework of the stream pipeline architecture, introducing a lifecycle state machine for lazy pipeline management, WebRTC latency optimizations, and supporting infrastructure.

## Review Progression

This PR is organized into 35 commits across 10 phases. Each phase builds on the previous one and can be reviewed as a unit.

### Phase 1: Dependencies and tooling (commits 1-4)

Cargo-only changes. Adds `gst-rtp` (RTP header extension support), `proptest` (lifecycle/property tests), and bumps `vergen-gix`. Lock file updated last.

### Phase 2: Small independent cleanups (commits 5-8)

Four self-contained fixes across unrelated modules (mavlink, zenoh, server, settings). No streaming behavior changes. Quick to review.

### Phase 3: CLI, helpers, and runtime setup (commits 9-11)

Adds the new CLI/runtime prerequisites: `--enable-dot`, `--rtsp-port`, `--disable-onvif`, `--enable-realtime-threads`, and the feature-gated `--enable-webrtc-task-test` compatibility hook, plus Linux thread-priority helpers and an explicit Tokio runtime builder replacing `#[tokio::main]`.

### Phase 4: Stream types and lifecycle foundation (commits 12-13)

The core addition: a lock-free atomic lifecycle state machine with four phases (`Idle`, `Waking`, `Running`, `Draining`) and consumer reference counting. `StreamStatusState` is added to the stream types module.

### Phase 5: GStreamer utility additions (commits 14-17)

New helpers that later sink and pipeline commits depend on: enum-nick property setting, element excision, proxysrc queue removal, jitterbuffer bypass, pipeline debug dumping, and hardened probe-pipeline teardown.

### Phase 6: RTSP server and sink rework (commits 18-19)

Replaces the `shmsrc`/`shmsink` shared-memory RTSP bridge with an `appsrc`/`appsink` in-process bridge. The RTSP server now receives encoded frames directly from the video tee. The RTSP sink gains valve-based flow control and PTS offset tracking across wake cycles.

### Phase 7: Sink adaptations (commits 20-23)

Adapts all sink types to the new architecture:
- `sink/mod.rs`: tighter unlink/EOS handling
- `udp_sink` / `zenoh_sink`: RTP queue removal + proxysrc queue excision for lower latency
- `webrtc_sink`: playout-delay extension, FEC/RED/RTX excision on connect, hardened teardown in `Drop`

### Phase 8: Pipeline module rework (commits 24-27)

Routes RTSP through `video_tee` instead of `rtp_tee`, adds `RawRtpTee` for source-info preservation in ONVIF/redirect pipelines, hardens the bus watcher, and enables pipeline debug dumps behind `--enable-dot`.

### Phase 9: Stream core and manager integration (commits 28-29)

Integrates the lifecycle state machine into `Stream` and rewrites the watcher loop from a simple restart loop to a four-phase lifecycle controller. The stream manager gets lifecycle-aware session handling and cooldown-based thumbnail behavior.

### Phase 10: WebRTC and rollout polish (commits 30-35)

Final integration work: WebRTC signalling/session cleanup, lower-priority TURN threads, BlueROV/test configuration updates, `/dot` endpoint gating, removal of the superseded WebRTC leak workflow, and the final module-level polish commit.

## Test plan

- [ ] Verify RTSP streaming works end-to-end
- [ ] Verify WebRTC session establishment and teardown
- [ ] Verify lazy idle: pipeline suspends after 5s with no consumers, resumes on demand
- [ ] Verify MavlinkCamera heartbeats persist across idle/wake cycles
- [ ] Verify no RTSP freeze after WebRTC client disconnection
- [ ] Verify thumbnail capture works during active streaming
- [ ] Measure WebRTC latency improvement vs master


patrickelectric (commented):
What is this point of this commit ? It appears unnecessary complexity with almost no trade-off. 
56d78ef74ef35f3bd43ace4fc8e3ed917855eaa3

patrickelectric (commented):
Zenoh pipeline should always work to make the video available for all services, since it's a pub/sub//signal/slot there is no way for us to identify if there is a client connected.
How will this work with zenoh ?
4dd6d3353b0960027852e3dab4daa9cbf29d2817

joaoantoniocardoso:
> How will this work with zenoh

UDP and Zenoh are always-on, so they never allow the stream to be idle.

src/lib/stream/lifecycle.rs
```diff
@@ -0,0 +1,975 @@
...
+    pub fn transition_to_running(&self) -> bool {
+        loop {
+            let cur = self.packed.load(Ordering::Acquire);
+            let (phase, count) = unpack(cur);
+            match phase {
+                Phase::Waking => {
+                    let new = if count > 0 {
+                        pack(Phase::Running, count)
+                    } else {
+                        pack(Phase::Draining, 0)
```
patrickelectric, line 175: Just to be sure, this moves walking -> draining, but returns true.
The function name is `transition_to_running`, and that's not the case.

src/lib/stream/lifecycle.rs
```diff
@@ -0,0 +1,975 @@
...
+        }
+
+        if phase == Phase::Draining {
+            assert_eq!(count, 0, "I3 violated: Draining with count={count}");
+        }
+
+        assert_ne!(
+            self.stream_status(),
+            StreamStatusState::Stopped,
+            "I5 violated: status returned Stopped"
```
patrickelectric: Is there a I4 ?

src/lib/stream/gst/utils.rs
```diff
@@ -197,6 +197,32 @@ pub fn wait_for_element_state(
...
+
+    loop {
+        std::thread::sleep(std::time::Duration::from_millis(polling_time_millis));
+
+        if element.current_state() == state {
+            break;
+        }
+
+        trials -= 1;
+        if trials == 0 {
```
patrickelectric: <=

src/lib/stream/pipeline/redirect_pipeline.rs
```diff
@@ -120,17 +135,22 @@ impl RedirectPipeline {
...
                 format!(
                     concat!(
-                        " ! rtph265depay",
+                        " ! tee name={raw_rtp_tee} allow-not-linked=true",
+                        " {raw_rtp_tee}.",
+                        " ! rtph265depay source-info=true",
                         " ! queue leaky=downstream silent=true flush-on-eos=true max-size-buffers=1 max-size-bytes=0 max-size-time=0",
                         " ! h265parse config-interval=-1",
-                        " ! capsfilter name={filter_name} caps=video/x-h265,profile={profile},stream-format=byte-stream,alignment=au",
+                        " ! capsfilter name={filter_name} caps=video/x-h265,stream-format=byte-stream,alignment=au",
```
patrickelectric, line 143: is there a reason for this to use byte-stream over avc in h264 ?

src/lib/stream/pipeline/redirect_pipeline.rs
```diff
@@ -120,17 +135,22 @@ impl RedirectPipeline {
...
                         " ! tee name={video_tee_name} allow-not-linked=true",
-                        " ! rtph265pay aggregate-mode=zero-latency config-interval=-1 pt=96",
-                        " ! tee name={rtp_tee_name} allow-not-linked=true"
+                        " {raw_rtp_tee}.",
+                        " ! rtph265depay source-info=true",
+                        " ! h265parse config-interval=-1",
```
patrickelectric, line 147: the other branch of the pipeline has a queue before h265parse

src/lib/stream/mod.rs
```diff
@@ -618,25 +879,60 @@ impl Drop for StreamState {
...
-        if let Err(error) =
-            wait_for_element_state(pipeline.downgrade(), ::gst::State::Null, 100, 10)
-        {
-            let _ = pipeline.set_state(::gst::State::Null);
-            error!("Failed setting Pipeline state to Null. Reason: {error:?}");
+
+        if let Some(handle) = eos_handle {
+            let _ = handle.join();
         }
+        let _ = pipeline_state;
```
patrickelectric: what ?

src/lib/stream/manager.rs
```diff
@@ -277,59 +280,146 @@ pub async fn get_jpeg_thumbnail_from_source(
...
+                if first_request {
+                    let cooldown_arc = cooldown.clone();
+                    let lifecycle_arc = lifecycle.clone();
+                    std::thread::Builder::new()
+                        .name("ThumbnailCooldown".into())
+                        .spawn(move || loop {
+                            std::thread::sleep(THUMBNAIL_COOLDOWN);
+                            let mut guard = cooldown_arc.lock().unwrap();
+                            match *guard {
+                                Some(last) if last.elapsed() >= THUMBNAIL_COOLDOWN => {
```
patrickelectric, line 411: why is this necessary if we already have the sleep ?

.github/workflows/test_webrtc_leak.yml
```diff
@@ -1,106 +0,0 @@
-name: Test webrtc thread leak
```
patrickelectric, line 1: I failed to follow where the leak detection is being done. And the inner infrastructure that this actions uses are still in the code.

joaoantoniocardoso: Yes, I had to decide where to split the full `next-18` branch changes and I put this change into this side:
* this test (falsely) fails for the new pipeline,
* and it's not worth fixing it twice.

On the following PR, the test is improved and it runs in the same CI workflow as all the other tests, against different GST versions.

You can check [this file from next PR](https://github.com/joaoantoniocardoso/mavlink-camera-manager/blob/2fc2ef896152eeb58a22c3928a193c3c569b439a/tests/integration/remote_integration/thread_leak.rs), which uses the new client instead of relying on the browser.

joaoantoniocardoso:
> What is this point of this commit ? It appears unnecessary complexity with almost no trade-off. [56d78ef](https://github.com/mavlink/mavlink-camera-manager/commit/56d78ef74ef35f3bd43ace4fc8e3ed917855eaa3)

* The `rtsp_port`, `disable_onvif` are required for testing and development.
* The `enable_realtime_threads` is for the convenience of developing without having to keep two binaries every time I want to compare the threading model.
* The `enable_dot` has a cost, I don't remember exactly what it was, but the performance was degraded when I had the developer page open and I traced down to this developer-targeted feature.
* The `stream_recreation_failure_timeout` was added to support the previous behavior of keep trying to recreate streams forever.

src/lib/stream/mod.rs
joaoantoniocardoso: those are dropping values/references

we can write `drop(pipeline_state);`

src/lib/stream/manager.rs
joaoantoniocardoso: The sleep ensures the minimum wait time, and the guarded code is the actual cooldown logic, which allows the cooldown to be 'refreshed' when another request is made within these x seconds.

src/lib/stream/pipeline/redirect_pipeline.rs
joaoantoniocardoso: yes, to decouple the video_tee branch from the main flow (rtp_tee). It's debatable whether it's really necessary, both in terms of threading and latency, but I kept it as original. Maybe we can remove it later.

src/lib/stream/pipeline/redirect_pipeline.rs
joaoantoniocardoso: The H264's `avc` equivalent for `H265` is the `hvc1`. 

This was a bit while ago, but from what I can remember, the `hvc1` had some incompatibility.

.github/workflows/test_webrtc_leak.yml
patrickelectric: > You can check [this file from next PR](https://github.com/joaoantoniocardoso/mavlink-camera-manager/blob/2fc2ef896152eeb58a22c3928a193c3c569b439a/tests/integration/remote_integration/thread_leak.rs), which uses the new client instead of relying on the browser.

An user will use a browser, not a rust based client.

src/lib/stream/mod.rs
patrickelectric: IIRC, pipeline_state is a reference, drop has no effect over it.

patrickelectric: Also, you are shadowing the variable soon, after, the drop will happens in either case, this is just redundant.

.github/workflows/test_webrtc_leak.yml
joaoantoniocardoso: Yes, but:
* the leaks are independent from client implementation. It's all about managing life-times of implicit contracts between our code, gstreamer and libnice.
* if we want to validate webrtc negotiation with a browser client, we can set up a proper end-to-end test for that.

patrickelectric:
can you squash the fixup commits ?

patrickelectric (approved):
image

PR #604, opened 2026-05-04
Fix segfault on ICE

joaoantoniocardoso:


src/lib/helper/tls.rs
```diff
@@ -0,0 +1,142 @@
...
+/// `gnutls_hmac_init()` fills HMAC handles correctly, and `libnice`'s
+/// STUN authentication works.
+///
+/// On systems where the original bug cannot occur, the call is
+/// harmless. If `glib-networking` is not installed, GIO returns a
+/// no-op fallback backend, and `libgnutls` (whose
+/// `_gnutls_global_init_skip` override is missing) is auto-initialized
+/// the normal way. If a non-GnuTLS backend is the default, GIO
+/// initializes that one instead, and `libgnutls` is either not loaded
+/// at all or, if `libnice` loads it, has no override stopping its own
```
patrickelectric, line 132: Are we sure of this comments or this is an AI guess ?

Cargo.toml
```diff
@@ -143,6 +144,6 @@ serde = { version = "1.0", features = ["derive"] }
 regex = "1"
 
 [profile.release]
-lto = true
-codegen-units = 1
-strip = true
+# lto = true
```
patrickelectric: Is this lost changes ?

patrickelectric (changes requested):

src/lib/helper/tls.rs
joaoantoniocardoso: Sure. This was line by line from a crash tracing dump

Cargo.toml
joaoantoniocardoso: Yes, oops

patrickelectric (approved):

PR #594, opened 2026-03-26
[next] [6] Add integration test suite with dynamic ports and parallel execution

joaoantoniocardoso:
## Summary

Adds the `clients` crate, a comprehensive integration test suite with dynamic port allocation, a latency comparison example, and CI updates for `nextest` plus multi-GStreamer compatibility coverage.

## Review Progression

The incremental part of this PR is organized into 22 commits across 6 phases.

### Phase 1: Existing source modifications (commits 1-2)

Two small source changes that the new client/test code depends on. Commit 1 adds `Deserialize` derives and a `to_nanos()` helper to `zenoh_message` types so the Zenoh client can deserialize frames. Commit 2 removes the old WebRTC leak-detection helper and its now-unused CLI hook because this PR replaces that path with broader browser/integration coverage.

### Phase 2: Cargo manifests (commits 3-5)

Cargo-only changes. Adds the new `clients` workspace member manifest, registers it in the workspace root together with test/example entries, then updates the lock file.

### Phase 3: `clients` crate (commits 6-13)

The new library crate, starting with the shared abstractions (`StreamClient`, `Codec`, frame probe helpers) and signalling protocol types, followed by one commit per client implementation: thumbnail, RTSP, UDP, Zenoh, and WebRTC. The final commit in this phase reorders module declarations.

### Phase 4: Integration test suite (commits 14-19)

The harness comes first: `McmProcess` with ephemeral ports, a REST client wrapper, and polling/state helpers. Then the test modules build outward in scope:
- `data_flow`: end-to-end verification for every sink type
- `lazy_lifecycle`: state transitions, idle timing, recovery, and per-sink lifecycle checks
- `mavlink_heartbeat`: heartbeat persistence through idle cycles
- `redirect_webrtc`: redirect pipelines combined with WebRTC/thumbnail/Zenoh consumers
- `remote_integration`: heavier fake/real camera scenarios, mixed concurrent consumers, and the disabled-ONVIF coverage that required small source-side guard rails

### Phase 5: Example (commit 20)

A latency measurement CLI tool that receives the same stream via RTSP, WebRTC, and UDP simultaneously, correlates frames by content hash, and computes pairwise latency deltas. Includes a companion Python plotting script.

### Phase 6: CI configuration (commits 21-22)

Moves CI from `cargo test` to `cargo nextest run`, adds `gstreamer1.0-plugins-bad`, introduces `.config/nextest.toml`, and adds the multi-GStreamer compatibility workflow.

## Test plan

[...]

PR #610, opened 2026-05-21
Restore decoupling

joaoantoniocardoso:
As a follow-up to some problems found in #593's introduced strategy of removing the decoupling between the sinks, this patch:

* Finds the missing piece that previously made the WebRTCSink queue accumulate buffers: WeBRTCBin's inner clocksync elements were sync=true. Now we finally have a truly sync=false WebRTC, which means it always runs with empty queues in steady state.
* Adds the decoupling back to each sink, so disturbances on one sink can't propagate upstream and/or to other sinks.
* Removes unneeded queues from the source pipelines so oscillations that amplify freeze-and-burst events can't happen.
* Refactors WebRTCSink so it resembles the other sinks more closely.
* The added measured latency sits between 2 and 5ms total (Pi 4), which seems a good price for the robustness.

Solves:
* https://github.com/bluerobotics/BlueOS/issues/3927

Evaluation:
* Tested against behavioral regressions (using #594).
* Tested against performance regressions using hypothesis tests.
* Tested against high CPU use and network impairment.


src/lib/stream/sink/udp_sink.rs
```diff
@@ -5,19 +5,17 @@ use gst::prelude::*;
 use tracing::*;
 
 use crate::{
-    stream::{gst::utils::excise_proxysrc_queue, pipeline::runner::PipelineRunner},
-    video_stream::types::VideoAndStreamInformation,
+    stream::pipeline::runner::PipelineRunner, video_stream::types::VideoAndStreamInformation,
 };
 
-use super::{SinkInterface, link_sink_to_tee, unlink_sink_from_tee};
+use super::{SinkInterface, link_sink_to_tee, make_proxy_bridge, unlink_sink_from_tee};
```
patrickelectric, line 11: `make_proxy_bridge` is added but not being used.

patrickelectric: This function is also being added in a later commit: c8415d2a1f8ca535beff3a3d124b6e3b1aa374d4

patrickelectric: and here: 4a02c209f6ff08e94b4219864e725db4d93eb4e5

src/lib/stream/pipeline/mod.rs
```diff
@@ -284,6 +284,11 @@ impl PipelineState {
             error!("Failed to syncronize children states. Reason: {error:?}");
         }
 
+        // Start the sink's own sub-pipeline runner
+        if let Err(error) = sink.start() {
+            warn!("Failed to start sink {sink_id}: {error:?}");
```
patrickelectric, line 289: just to be sure, we are failing here, but the insert will happen still ?

patrickelectric (changes requested):

src/lib/stream/pipeline/mod.rs
joaoantoniocardoso: Great question. In short, this is keeping the same behavior, which contains a fragility we need to improve. See below.

Each Pipeline has a PipelineRunner, which runs independently. The "Streams" are groups of _one Source Pipeline + multiple Sink Pipelines_.

When the _Source Pipeline_ of a Stream fails (e.g., if v4l2src's device is closed by Linux), there's a watcher that tears down and recreates the whole Stream (Source Pipeline + each Sink Pipeline).

When a _Sink Pipeline_ of a Stream fails (e.g., if something happens to the socket), it just dies forever.

The lack of communication upstream (from Sink Pipelines to Source Pipelines) was intentional: we don't want the Source or other Sinks to be disturbed by a Sink. What it is lacking is the same recreation strategy used by the Source Pipeline.

Also, in the future, the API should surface the status of each Pipeline Sink of a Stream, but for that, we must semantically separate the Stream concept from the Source Pipeline concept, which today is _very_ liquid/mixed because that _was_ the original design.

patrickelectric: Ok, that's a pretty fair point.

patrickelectric (approved):

PR #615, opened 2026-07-09
Improve gh CI caching

joaoantoniocardoso:


patrickelectric (commented):
I'm not following, CACHE_SAVE is only true if ref is master, but there is an action that runs for closed pull requests ? how will this work ?

joaoantoniocardoso:
> I'm not following, CACHE_SAVE is only true if ref is master, but there is an action that runs for closed pull requests ? how will this work ?

There are other caching that ends up being saved on the GH caching, so that's why I put the action to delete caches from closed PRs.

Ah, and the normal flow is: branches read the chache from master, and only master writes cache.

patrickelectric (approved):

PR #621, opened 2026-08-12
.github: workflows: Install apt deps without broken package cache

joaoantoniocardoso:
## Summary
- The `test` job fails on current PRs because `awalsh128/cache-apt-pkgs-action` restores a corrupt empty cache (~12KB), reports `Restoring 0 packages from cache`, and never installs `libglib2.0-dev` / GStreamer headers.
- Clippy then dies with `The system library glib-2.0 required by crate glib-sys was not found`.
- Replace the apt cache action with plain `apt-get install` for build and runtime deps.

Unrelated flake also seen on PRs: linux-desktop checkout `server certificate verification failed` — not addressed here.

## Test plan
- [ ] CI `test` job on this PR installs packages and gets past clippy/`glib-sys`
- [ ] CI `test` job runs `cargo test`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -70,9 +70,9 @@ jobs:
           path: frontend
 
       - name: Install build dependencies
-        uses: awalsh128/cache-apt-pkgs-action@v1
-        with:
-          packages: |
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends \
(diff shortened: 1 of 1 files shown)


joaoantoniocardoso:
Closing: preferred fix is deleting the broken apt package caches via the GitHub API, not replacing cache-apt-pkgs-action.

PR #626, opened 2026-08-21
lib: cli: manager: Expand string arguments

### LINKED ISSUE #619: All cmdline arguments should be shell-expanded.
~~This will allow passing `--mavlink-system-id=$MAV_SYSTEM_ID` in BlueOS, unlocking https://github.com/bluerobotics/BlueOS/issues/2396.~~ -> already done because we expand it before creating the shell cmdline.

Nevertheless, this is still a good feature.

patrickelectric:
Expand env vars in argv before clap parse, same as blueos-recorder.
log_path, settings_file and zenoh_config_file still use shellexpand::tilde for ~.
TURN servers can also be passed as MCM_TURN_SERVERS.
turn: and onvif: URLs without a host now fail at startup.

Fix #619

### DIFF
--- a/src/lib/cli/manager.rs
+++ b/src/lib/cli/manager.rs
@@ -1,4 +1,4 @@
-use std::{collections::HashMap, sync::Arc, time::Duration};
+use std::{collections::HashMap, ffi::OsString, sync::Arc, time::Duration};
 
 use anyhow::Context;
 use clap;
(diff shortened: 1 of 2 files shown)


joaoantoniocardoso:
I used a clap's "expanded" somewhere in another project... Can you try finding/using the feature? It'd simplify things.

edit: was in blueos-recorder:

```rust
/// Constructs our manager, Should be done inside main
pub fn init() {
    let expanded_args = std::env::args()
        .map(|arg| {
            // Fallback to the original if it fails to expand
            shellexpand::env(&arg.clone())
                .inspect_err(
                    |_| warn!(arg = ?arg, "Failed expanding arg, using the non-expanded instead"),
                )
                .unwrap_or_else(|_| arg.into())
                .into_owned()
        })
        .collect::<Vec<String>>();

    let reparsed_expanded_args = Args::parse_from(expanded_args);

    init_with(reparsed_expanded_args);
}
```

patrickelectric:
done, using the same expand before parse

src/lib/cli/manager.rs
```diff
@@ -192,12 +197,32 @@ lazy_static! {
...
+                clap_matches: Args::parse_from(["mavlink-camera-manager"]),
+                expand_errors: Vec::new(),
+            };
+        }
+
+        let mut expand_errors = Vec::new();
+        let expanded_args = std::env::args_os()
+            .enumerate()
+            .map(|(index, arg)| match arg.into_string() {
+                Ok(arg) => match shellexpand::env(&arg) {
```
joaoantoniocardoso: now I'm wondering whether we should expand::env or expand::full here ; any thoughts?

patrickelectric: env, same as recorder. `full` would also expand `~` in every argv, including turn/onvif urls.
log_path, settings_file and zenoh_config_file still use tilde.

patrickelectric:
ping @joaoantoniocardoso

joaoantoniocardoso (approved):

PR #629, opened 2026-08-22
src: lib: stream: Handle force-non-compliant-url

### LINKED ISSUE #460: Redirect stream: some cameras require `force-non-compliant-url=true` in order to work
Gstreamer fails to negotiate without it on the rtspsrc. ffmpeg and vlc can happily play the stream. I built a version.

[this](https://github.com/mavlink/mavlink-camera-manager/compare/master...Williangalvani:mavlink-camera-manager:master?expand=1) was enough for getting the camera stream to work, but I'm afraid it could break other cameras, so perhaps we should instead update `get_encode_from_stream_uri()` so it tries both ways?

patrickelectric:
Some cameras fail RTSP negotiation unless force-non-compliant-url=true, but hardcoding it breaks others.

Fix #460


joaoantoniocardoso (changes requested):
Hi,

I believe I forgot to document it in the issue, but we want to attack this with a strategy pattern, similar to how we handle the encoding: when it fails without it, then we try with this parameter. Once it succeeds the test with this parameter, it uses it on the actual stream.

PR #630, opened 2026-08-22
ci: Fix Windows GStreamer MSI download

patrickelectric:
gstreamer.freedesktop.org returns 503 during the MSI download.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -210,6 +210,15 @@ jobs:
         run: |
           brew install gstreamer gst-plugins-base gst-plugins-good gst-plugins-bad gst-libav gst-rtsp-server
 
+      - name: Cache GStreamer MSIs
+        if: ${{ contains(matrix.TARGET, 'windows') }}
+        uses: actions/cache@v6.1.0
+        with:
+          path: |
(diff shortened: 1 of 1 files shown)


patrickelectric:
Windows CI still red because `gstreamer.freedesktop.org` is broken, not because of this change.

What this PR does:
- curl with `--fail` + retries (no more silent 515-byte HTML "MSIs")
- reject MSI files under 1MB
- check msiexec exit code
- require `pkg-config.exe` after install
- cache the MSIs by version

What we hit on runners (and locally):
- GitHub Actions (IAD): HTTP 503 from Varnish for the 1.18.6 MSVC MSIs
- Direct download: same 503 from several networks
- When a Varnish HIT appears, the transfer truncates every time (~91M of 106M runtime, ~14M of 169M devel) with `curl: (18) end of response with N bytes missing`

So we cannot mirror a good copy onto a fork release until freenode's CDN recovers (or someone uploads known-good MSIs from an old local install).

Until then: non-Windows jobs are the signal; Windows will keep failing at Install GStreamer on download.

joaoantoniocardoso:
It seems that a big portion of their infrastructure is down, including the homepage, so they are probably working on it.

If we don't want to depend on that, we should seek a mirror or something like Chocolatey.

joaoantoniocardoso:
Their infra is back; CI is passing on master.

PR #631, opened 2026-08-23
fix(stream/manager): recover cooldown flag on error and ensure panic-safe cleanup

ARYA-mgc:
### What this fixes

1. **Permanent wedging on `add_consumer()` failure**:
   If `lifecycle.add_consumer().await` failed on an initial thumbnail request, `thumbnail_cooldown` was left set to `Some(...)`. Every subsequent request skipped waking the stream and timed out after 30 seconds indefinitely.
   - **Fix**: Reset `thumbnail_cooldown` to `None` on error so future requests can retry.

2. **Panic-safe cleanup**:
   If a panic occurred before the cooldown thread took ownership, the consumer was never removed, permanently pinning the stream active.
   - **Fix**: Added a `scopeguard` (already in workspace from `sink/mod.rs`) to guarantee consumer removal and cooldown reset on panic.

3. **Thread spawn error & poison resilience**:
   - If `thread::spawn` fails under thread exhaustion, cleanup runs immediately instead of leaking the consumer via `.ok()`.
   - Handled poisoned mutex locks safely with `unwrap_or_else(PoisonError::into_inner)`.

### Architecture Note
- Dedicated OS thread + Tokio runtime isolation for GStreamer thumbnails remains fully intact.

### DIFF
--- a/src/lib/stream/manager.rs
+++ b/src/lib/stream/manager.rs
@@ -448,7 +448,7 @@ pub async fn get_jpeg_thumbnail_from_source(
                 if first_request {
                     let cooldown_arc = cooldown.clone();
                     let lifecycle_arc = lifecycle.clone();
-                    std::thread::Builder::new()
+                    let spawn_result = std::thread::Builder::new()
                         .name("ThumbnailCooldown".into())
                         .spawn(move || {
                             loop {
@@ -465,8 +465,16 @@ pub async fn get_jpeg_thumbnail_from_source(
                                     _ => {}
                                 }
                             }
-                        })
-                        .ok();
+                        });
+
+                    if let Err(error) = spawn_result {
+                        warn!("Failed to spawn thumbnail cooldown thread: {error}");
+                        let mut guard = cooldown
+                            .lock()
+                            .unwrap_or_else(std::sync::PoisonError::into_inner);
+                        *guard = None;
+                        lifecycle.remove_consumer_in_background();
+                    }
                 }
 
                 let _ = tx.send(res);
--- a/src/lib/stream/manager.rs
+++ b/src/lib/stream/manager.rs
@@ -356,6 +356,20 @@ pub async fn get_jpeg_thumbnail_from_source(
                     false
                 };
 
+                // Panic-safe guard: if anything panics before the cooldown thread
+                // takes ownership, ensure the consumer is removed and cooldown is reset.
+                let mut consumer_guard = first_request.then(|| {
+                    let lifecycle_bg = lifecycle.clone();
+                    let cooldown_bg = cooldown.clone();
+                    scopeguard::guard((), move |()| {
+                        let mut g = cooldown_bg
+                            .lock()
+                            .unwrap_or_else(std::sync::PoisonError::into_inner);
+                        *g = None;
+                        lifecycle_bg.remove_consumer_in_background();
+                    })
+                });
+
                 // Wait for the pipeline to be alive and Playing.
                 // Truly-idle streams need full Waking (pipeline recreation +
                 // position advance). Draining/Waking/Running streams may
@@ -377,6 +391,9 @@ pub async fn get_jpeg_thumbnail_from_source(
                                         "Failed to remove thumbnail consumer after timeout: {error}"
                                     );
                                 }
+                            }
+                            if let Some(guard) = consumer_guard.take() {
+                                scopeguard::ScopeGuard::into_inner(guard);
                             }
                             let _ = tx.send(Some(Err(Arc::new(anyhow!(
                                 "Pipeline did not resume in time for thumbnail"
@@ -474,6 +491,12 @@ pub async fn get_jpeg_thumbnail_from_source(
                             .unwrap_or_else(std::sync::PoisonError::into_inner);
                         *guard = None;
                         lifecycle.remove_consumer_in_background();
+                    }
+
+                    // Defuse the panic guard — either the cooldown thread has taken
+                    // ownership or we cleaned up on spawn error.
+                    if let Some(guard) = consumer_guard.take() {
+                        scopeguard::ScopeGuard::into_inner(guard);
                     }
                 }
 
--- a/src/lib/stream/manager.rs
+++ b/src/lib/stream/manager.rs
@@ -331,7 +331,9 @@ pub async fn get_jpeg_thumbnail_from_source(
                 // spawn the cooldown cleanup thread later.
                 let first_request;
                 {
-                    let mut guard = cooldown.lock().unwrap();
+                    let mut guard = cooldown
+                        .lock()
+                        .unwrap_or_else(std::sync::PoisonError::into_inner);
                     first_request = guard.is_none();
                     *guard = Some(std::time::Instant::now());
                 }
@@ -383,7 +385,9 @@ pub async fn get_jpeg_thumbnail_from_source(
                             debug!("Pipeline did not resume in time for thumbnail");
                             if first_request {
                                 {
-                                    let mut guard = cooldown.lock().unwrap();
+                                    let mut guard = cooldown
+                                        .lock()
+                                        .unwrap_or_else(std::sync::PoisonError::into_inner);
                                     *guard = None;
                                 }
                                 if let Err(error) = lifecycle.remove_consumer().await {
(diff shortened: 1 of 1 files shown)


ARYA-mgc:
@patrickelectric sorry for the direct tag - could you review this when you get a chance?

joaoantoniocardoso:
Hi @ARYA-mgc, 

I am unable to reproduce the thread leak.

Could you write down the step-by-step on how to reproduce it?

Add details like the platform and GStreamer version you are testing on.

Thanks

ARYA-mgc:
Hi @joaoantoniocardoso , sorry for the delay - I was tied up with some personal stuff, but I'm back on this now!

**Platform:** Ubuntu 22.04, GStreamer 1.24.x

### Reproducing the thread churn

```bash
# Watch thread count
watch -n 0.5 "ls /proc/$(pgrep mavlink-camera)/task | wc -l"

# Fire 50 concurrent requests (different quality values bypass the cache)
for i in $(seq 1 50); do
  curl -s "http://localhost:6020/thumbnail?source=/dev/video0&quality=$i" -o /dev/null &
done
```

Thread count spikes by ~50 because each request spawns an OS thread + Tokio runtime. They do exit after finishing, but under load they all exist at once.

### The bigger issue (Bug 4)

If `add_consumer()` ever fails, the cooldown flag is never cleared. Every subsequent thumbnail request skips `add_consumer()`, waits 30s, and times out -  permanently. The fix adds `cooldown.lock().unwrap().take()` on the error path.

### Other fixes

- Cooldown thread replaced with `tokio::spawn` (no more blocking OS thread for 15s)
- `scopeguard` added so a panic can't leave a stream stuck alive forever

Let me know if you'd like me to split these into separate PRs!

joaoantoniocardoso:
> Let me know if you'd like me to split these into separate PRs!

Yes, if they don't depend on each other.

---

> Reproducing the thread churn

I used your commands and could not reproduce any leaks:
1. Created a fake RTSP source on localhost; there were 20 threads.
2. Launched your command; threads grew to 49.
3. Waited a few seconds; threads got back to the same initial 20.

I'm now assuming you did **not** mean a thread _leak_ claim at all, but a performance claim?

The patch is reducing the pile of threads but also reducing the runtime isolation, which may have been overkill, but was intentionally added to reduce interference from the thumbnail workload.

To be transparent with you, my time is a bit scarce now, so it would be helpful if you could explain the underlying real-world context, the use case, and the real pain so we can better reason and discuss the trade-offs.

Thanks

ARYA-mgc:
Hi @joaoantoniocardoso, thanks for testing this and for the clarification.

You’re right about the terminology: this is **thread churn / peak thread concurrency**, rather than a persistent thread leak. My original wording was inaccurate.

The real-world motivation came from running `mavlink-camera-manager` on resource-constrained companion computers, where bursts of thumbnail requests can create significant temporary thread/runtime overhead and CPU scheduling pressure. I also understand the reason for keeping the thumbnail workload isolated from the main runtime.

Following your suggestion, I’ll split the changes:

1. **This PR (#631): focused bug fixes only**

   * Keep the existing dedicated OS thread + Tokio runtime architecture unchanged.
   * Fix the `add_consumer()` failure path so `cooldown` is reset when consumer creation fails, preventing subsequent requests from waiting 30 seconds on a pipeline that was never started.
   * Ensure consumer cleanup happens reliably on error paths.

2. **Follow-up PR / RFC: thread-model optimization**

   * Address the thread churn separately while preserving the runtime isolation that was intentionally introduced.
   * Explore bounded/constrained approaches rather than changing the architecture in this PR.

I’ll update the branch with the scoped-down changes by tonight or tomorrow morning (IST) and keep the performance/thread-model discussion separate.

Thanks again for pointing this out.

ARYA-mgc:
Hi @joaoantoniocardoso,

I've updated the branch with the scoped-down changes:

1. **Dedicated OS thread + Tokio runtime isolation remains completely unchanged.**
2. **Fixed the permanent wedging bug**: If `add_consumer()` fails on an initial thumbnail request, the `thumbnail_cooldown` flag is now properly reset so future requests can retry rather than perpetually timing out after 30 seconds.
3. **Panic & error cleanup**: Guarded consumer removal with `scopeguard` (already used in `sink/mod.rs`) and added cleanup if `thread::spawn` fails.
4. **Poison resilience**: Handled poisoned lock recovery across `thumbnail_cooldown`.

I've also updated the PR title and description to reflect this focused scope. Whenever you have a moment, please take another look!

joaoantoniocardoso:
Hi @ARYA-mgc,

Nice, can you split that into 3~4 commits according to the 4 points you explained?

Also, can you give a step-by-step how to reproduce the problem?

Thanks!

ARYA-mgc:
Hi @joaoantoniocardoso,

Done! I rebased everything into 4 atomic commits:

1. **Reset cooldown on `add_consumer()` failure** - fixes the main wedging bug.
2. **Handle `thread::spawn()` failure** - cleans up properly if spawning fails.
3. **Add `scopeguard` cleanup** - ensures cleanup even if a panic happens.
4. **Handle poisoned mutex** -avoids `.lock().unwrap()` for `thumbnail_cooldown`.

The main bug was that when `add_consumer()` failed, `thumbnail_cooldown` stayed set, so future requests skipped `add_consumer()` and kept timing out.

Commits 2–4 are just defensive hardening for edge cases.

ARYA-mgc:
Hi @joaoantoniocardoso,

Here are the step-by-step reproduction instructions for the `thumbnail_cooldown` failure case.

## Bug: `thumbnail_cooldown` remains set after `add_consumer()` fails

### Root cause

In the original implementation, `thumbnail_cooldown` is set before calling `lifecycle.add_consumer()`.

If `add_consumer()` returns an error, the function returns without clearing the cooldown:

```rust
let first_request;
{
    let mut guard = cooldown.lock().unwrap();
    first_request = guard.is_none();
    *guard = Some(std::time::Instant::now());
}

let need_position_advance = if first_request {
    match lifecycle.add_consumer().await {
        Ok(snapshot) => {
            // ...
        }
        Err(error) => {
            // cooldown remains Some(...)
            let _ = tx.send(Some(Err(Arc::new(error))));
            return;
        }
    }
} else {
    false
};
```

The next thumbnail request therefore sees `first_request == false`, skips `add_consumer()`, and waits for the existing cooldown/pipeline state. In the failure case this can result in repeated 30-second timeouts.

### Reproduction

**Platform:** Ubuntu 22.04
**GStreamer:** 1.24.x

Start the camera manager:

```bash
mavlink-camera-manager --verbose
```

Use a valid stream/source for the thumbnail request.

To make the race deterministic, temporarily add a delay immediately after releasing the cooldown lock and before `add_consumer()`:

```rust
{
    let mut guard = cooldown.lock().unwrap();
    first_request = guard.is_none();
    *guard = Some(std::time::Instant::now());
}

tokio::time::sleep(tokio::time::Duration::from_secs(2)).await;

let need_position_advance = if first_request {
    match lifecycle.add_consumer().await {
```

Then:

**Terminal A**

Start a thumbnail request:

```bash
curl -s "http://localhost:6020/thumbnail?source=/dev/video0&quality=85" -o /dev/null
```

While the request is inside the temporary 2-second delay:

**Terminal B**

Remove the corresponding stream:

```bash
curl -X DELETE "http://localhost:6020/streams/<stream-id>"
```

The purpose of this step is to make `add_consumer()` fail after `thumbnail_cooldown` has already been set.

After restoring/re-creating the stream, send another thumbnail request:

```bash
time curl -s "http://localhost:6020/thumbnail?source=/dev/video0&quality=85" -o /dev/null
```

With the original code, the failed first request leaves `thumbnail_cooldown` populated. Subsequent requests therefore do not retry `add_consumer()` and can repeatedly hit the 30-second timeout.

### Verification

With commit `e4f6e33`, the error path clears the cooldown:

```rust
Err(error) => {
    let mut guard = cooldown
        .lock()
        .unwrap_or_else(std::sync::PoisonError::into_inner);

    *guard = None;

    let _ = tx.send(Some(Err(Arc::new(error))));
    return;
}
```

After the same failure, the next request is able to enter the first-request path and retry `add_consumer()` instead of remaining wedged.

The important difference is that the original code leaves the cooldown set after the failed `add_consumer()` call, while the fix explicitly clears it.

